PR #4, opened 2016-06-05
changes to make color switching work (fixes handicap bug?)

thouis:
I think this fixes the handicap color-switching bug.


anoek:
Thanks for the patch!

PR #9, opened 2017-01-27
initial support to pass clock info to bot

roy7:


anoek:
Please don't say "Update gtp2ogs.js" for the description of your commits, provide a description

anoek:
One thing we'll need for this is a command line option to disable this, and/or the option to override / set an upper bound for how much time to allow a bot to think (which would be ideal). For official site bots ideally we'd like to set an upper bound of how much time a bot can spend on a certain move, so for blitz games it'll use the lower limit of what is remaining on the clock, but for long games we limit by some upper bound so our hard time limits don't kick in and kill the process.

gtp2ogs.js
```diff
@@ -305,6 +344,16 @@ class Game {
             this.state = gamedata;
             check_for_move();
         });
+        this.socket.on('game/' + game_id + '/clock', (clock) => {
```
anoek: This socket subscription can't live here, it needs to live with the other `socket.on` calls in the `Connection` instance as `loadState` gets called whenever a move is needing to be generated

roy7:
Going to make some revisions based on Leela author feedback and:

https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html

which is more detailed than the GnuGO GTP specification info. Specifically, if byoyomi > 0 and stones = 0 in time_settings, I was telling the bot time is unlimited. I've requested clarification from Leela's author on passing stone data via time_left in non-Canadian time controls.

gtp2ogs.js
roy7: I've been working on this but I don't fully understand why the clock subscription isn't in here with the other Game-specific subscriptions like gamedata, phase, and move.

roy7: Perhaps leave it where it is , but don't run loadClock(), simply update this.state.clock so the next loadState( ) has the info handy?

roy7:
Setting an upper bound on time bot will think might not be easily done via the time controls. There's no guarantee bots even read the time control data we provide them. For official bots where you want an upper limit I'd suggest running them with whatever command line option will cap the # of rollouts. Then the bot will use it's normal time management if it has any, but won't go past whatever hard cap you want for games with large timers.

gtp2ogs.js
anoek: My bad, this is fine here, I think I didn't click the code expandy button enough and thought this was living within the loadState call, which it most certainly is not. Code read fail.

gtp2ogs.js
```diff
@@ -169,11 +180,160 @@ class Bot {
...
+                    )
+                    + " " + state.time_control.periods);
+                this.command("time_left black " + black_timeleft + " " + (black_timeleft > 0 ? "0" : state.clock.black_time.periods));
+                this.command("time_left white " + white_timeleft + " " + (white_timeleft > 0 ? "0" : state.clock.white_time.periods));
+            } else {
+                // OGS enforces the number of periods is always 1 or greater. Let's pretend the final period is a Canadian Byoyomi of 1 stone.
+                // This lets the bot know it can use the full period per move, not try to fit the rest of the game into the time left.
+                //
+                let black_timeleft = Math.max( Math.floor(state.clock.black_time.thinking_time
+                    - black_offset + (state.clock.black_time.periods - 1) * state.time_control.periods), 0);
```
anoek: Shouldn't this be `(state.clock.black_time.periods - 1) * state.time_control.periods_time)` ?

gtp2ogs.js
```diff
@@ -169,11 +180,160 @@ class Bot {
...
+                let white_timeleft = Math.max( Math.floor(state.clock.white_time.thinking_time
+                    - white_offset + (state.clock.white_time.periods - 1) * state.time_control.periods), 0);
```
anoek: Ditto

gtp2ogs.js
```diff
@@ -169,11 +180,160 @@ class Bot {
...
+            }
+
+            this.command("time_left black " + (black_timeleft > 0 ? black_timeleft + " 0"
+                : Math.floor(state.clock.black_time.block_time - black_offset) + " " + state.clock.black_time.moves_left));
+            this.command("time_left white " + (white_timeleft > 0 ? white_timeleft + " 0"
+                : Math.floor(state.clock.white_time.block_time - white_offset) + " " + state.clock.white_time.moves_left));
+        } else if (state.time_control.system == 'fischer') {
+            // Not supported by kgs-time_settings and I assume most bots. So lets just handle it like absolute time. A bot may move quicker
+            // than it needs to when low on time, thinking this is total time left for whole game, but since time is added
+            // each move it should be somewhat self-correcting.
```
anoek: If this is causing the bots to think they only have such low time for the rest of the game, then maybe it would be better to treat it as byo-yomi with one period left of `time_increment` seconds, and main time being the time that's left minus `time_increment` ?

gtp2ogs.js
roy7: Haha. Um... yes...

gtp2ogs.js
roy7: Good idea!

roy7:
Server issues led to a timeout possible problem. If clock event comes in a long time before the move/active_game, the clock date told to bot is already seconds too old to be accurate since clock kept ticking while waiting for move to arrive.

Should store the clock data instead of telling bot immediately. Then when we're going to genmove, send the clock data at that time (based on now() not the now() when clock came in previously).

I saw this with a 5 second lag buffer but 7 seconds between clock event and move event, led to bot trying to think 2 seconds too long and losing a Fischer game. Hiccups like this only lose a period in byoyomi but Fischer has no periods. So if instead we updated our state, then calculated time remaining when we did the genmove, bot would have been told a time left 7 seconds smaller and all is well.

Sending clock info instantly before genmove might also let Leela use pondering (any input during pondering stops the pondering), since we'd leave bot alone until we're ready to genmove again.

PR #11, opened 2017-01-30
clearTimeout should be called, not clearInterval. Add --timeout option.

roy7:
Per game_id timeouts are set with setTimeout, not setInterval. This might be why after 10 minutes I see a Disconnect message after every move in the log. Google tells me some browsers use these interchangeably but they aren't supposed to be?


anoek:
:+1:

PR #12, opened 2017-02-01
Fix bug in timeout code. Disconnect when game is finished.

roy7:
active_game.game_id doesn't exist, it is named active_game.id.

We may as well disconnect as soon as a game is finished.

### DIFF
--- a/gtp2ogs.js
+++ b/gtp2ogs.js
@@ -308,7 +308,7 @@ class Game {
             if (!this.connected) return;
             this.log("gamedata")
 
-            //this.log("Gamedata: ", gamedata);
+            //this.log("Gamedata:", JSON.stringify(gamedata, null, 4));
             this.state = gamedata;
             check_for_move();
         });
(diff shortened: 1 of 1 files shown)


gtp2ogs.js
```diff
@@ -520,14 +523,17 @@ class Connection {
             if (gamedata.phase == "play" && gamedata.player_to_move == this.bot_id) {
                 this.processMove(gamedata);
             }
-
-
-            if (this.connected_game_timeouts[gamedata.game_id]) {
-                clearTimeout(this.connected_game_timeouts[gamedata.game_id])
+            if (gamedata.phase == "finished") {
+                this.disconnectFromGame(gamedata.id);
```
anoek: I think we need to potentially do a `clearTimeout` here to clear the timer that gets created in the else block

roy7: A clearTimeout is called within this.disconnectFromGame(). That should catch and clear it?

anoek: Right you are!

PR #14, opened 2017-02-01
Stone removal processing obsolete, server autoscores bot games. Minor typo fix in phase logging

roy7:


roy7:
With the old code in here, the bot was being called a second time for a fresh genmove when phase changed to stone removal.

PR #16, opened 2017-02-03
Initial support for bot process to persist between turns

roy7:
Here's the start towards optional persistent bot process. It worked in my testing pretty well with Leela. Two issues for further thought are:

1. Should .disconnect() kill the bot process? Are there ever times we .disconnect() and still want the bot to keep working? Clearly if the game has finished we can disconnect and kill the process. Currently only disconnect.
2. After bot passes, the bot is called to genmove a second time in a row. Maybe something to do with the pass and restart crash test, need to examine that further.

I'm doing the pull request so you can look over the diff and provide any feedback you might have as I keep working on it?

### DIFF
--- a/gtp2ogs.js
+++ b/gtp2ogs.js
@@ -491,6 +491,8 @@ class Game {
         this.bot = null;
         this.my_color = null;
 
+        // check_for_move is no longer called anywhere, now that moves are processed after gamedata and move messages from server
+        //
         let check_for_move = () => {
             if (!this.state) {
                 console.error('Gamedata not received yet for game, but check_for_move has been called');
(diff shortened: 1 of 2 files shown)


gtp2ogs.js
```diff
@@ -306,7 +306,7 @@ class Game {
                 if (this.waiting_on_gamedata_to_make_move && this.state.moves.length == this.move_number_were_waiting_for) {
                     this.makeMove(this.move_number_were_waiting_for);
                     this.waiting_on_gamedata_to_make_move = false;
-                    this.move_number_were_waiting_for = -1;
+                    this.move_number_were_waiting_for += 2;
                 }
```
roy7: By keeping the move we're waiting for current at all times, persistent bots can receive move events properly even if the active_game event comes in after the move.

roy7:
One issue I haven't tried to fix yet, if server resets and sends out a gamedata event instead of a move event, persistent bot doesn't know to send the move to the bot and so will try to genmove without knowing about prior opponent move. (Or possibly it thinks it changed colors, I'm not certain.)

Need to detect and handle this more cleanly (store # of last move we sent out?) or just kill and restart bot when an expected gamedata happens? Maybe trying to predict and correct corrupt states is too risky vs just restarting bot.

anoek:
I'd personally just restart the bot as that's a nice safe nuclear solution, but if we can reliably handle things so we don't have to do that, then that's cool too

roy7:
Still working on the conflicts, will comment when all set.

roy7:
I think merge conflicts are clean now. You should give all the code a fresh look over just in case.

roy7:
Ok the makeMove commit is the big one. I tested it on beta with restarting bot during my turn, bot's turn, cancelling game, and other basics. All seems to work. Since active_game is basically not used for anything now, there should be no more race conditions/etc. A 2nd gamedata will kill bot and restart on next move. New moves are requested immediately when a gamedata or move message arrives from server. That will also eliminate the bot losing time waiting for the active_game (only an issue on live, I assume temporarily).

check_for_move is no longer used. If you like I can comment all of that out, and all of the related code for waiting_on_gamedata_to_make_move and move_number_were_waiting_for. None of that gets checked/used currently. Or we can leave it be for time being.

I don't make any specific checks about game connection status, I assume that is handled automatically during initial connect to the game prior to when we get the first gamedata packet.

If the server disconnects us (totally, not just one game) we should do an immediate reconnect. Otherwise we wait for the setInterval to check for active_games, and can timeout on blitz matches.

anoek:
Lets just remove it all together if we're not going to use it, no sense in having dead code laying about.. if we need to look at it for some reason in the future, that's what git is for :)

roy7:
Ok I'll do that this evening and play a few test games. I'll simply remove all of the old code instead of commenting it out. :)

roy7:
Sweet!

PR #17, opened 2017-02-06
Refresh timeout when moves are made. Oops!

roy7:


roy7:
I added some more checks for timeouts to be sure they are all cleared when needed. It's possible if we're disconnected from a game but had a timeout set, we wouldn't clear the timeout... at least maybe. This way we're certain there's no loose timeout.

PR #23, opened 2017-03-02
Calculate clocks before genmove, not when we get a clock event

roy7:
Sometimes there is a long delay between the clock message and the move message. This time gets "lost" and the bot doesn't know about it, which can lead to a timeout. By calculating the loadClock() right before a genmove, which calls now() at that time instead of when the clock packet arrived, the lost time is subtracted from the bot's available time left.

This should be a huge improvement on reducing bot timeouts?


anoek:
Makes sense to me!

PR #26, opened 2017-03-12
Add a bunch more command line options to configure bot

roy7:
I will likely add some more beyond this, but wanted to go ahead and make these available for review/comment.

### DIFF
--- a/gtp2ogs.js
+++ b/gtp2ogs.js
@@ -7,7 +7,7 @@ let DEBUG = false;
 let PERSIST = false;
 let KGSTIME = false;
 let NOCLOCK = false;
-let REJECTALL = false;
+let REJECTNEW = false;
 
 let spawn = require('child_process').spawn;
 let os = require('os')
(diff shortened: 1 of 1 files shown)


roy7:
--maxrankedhandicap 0 is failing for me rarely, because sometimes a match has notification.handicap = -1. How is that possible? One such game is https://online-go.com/game/8057801. It is a ranked 9 handicap match, Leela is white. I was logging this to debug:

`conn_log("Testing", notification.handicap, " > ", argv.maxrankedhandicap);
if ( (argv.maxrankedhandicap !== undefined) && notification.ranked && (notification.handicap > argv.maxrankedhandicap) ) {
conn_log("Max ranked handicap is " + argv.maxrankedhandicap);
reject = true;
}
`

And it printed "Testing -1 > 0" for the 9 handicap game that was accepted. When I send my own 9 handicap matches to RoyalLeela, I see "Testing 9 > 0".

roy7:
Mystery solved, handicap = -1 means "automatic". Is that intended? I assumed automatic would choose the handicap when the challenge when out based on the ranks at the time of challenge. I guess I'd need to test the difference in ranks to support a mixture of automatic handicap and maxrankedhandicap settings?

anoek:
Yep -1 means automatic, I think we left it like that for two reasons: it makes it consistent when talking about open challenges vs direct challenges, and the second reason being that people's ranks can change between when a challenge was issued and when it is accepted, so we resolve what the handicap is at game creation time instead of challenge creation time.

roy7:
@anoek From the bot's perspective though, how would I know what the resulting handicap will actually be? To honor things like --maxrankedhandicap. I'd need to calculate the difference and accept/reject accordingly, but my own ranking isn't listed in the notification event. Should it? It's also going to be game speed specific isn't it?

PR #42, opened 2017-04-27
Catch unhandled errors and log them, reset connection

roy7:
Instead of conn.connection_reset() should we perhaps delete the connection and open a brand new one?

I sometimes get ECONNRESET which will kill the gtp2ogs process without this handler. I'm not sure if that is when the server restarts, or what exactly causes it. When running under Linux anyway, never seemed to have this issue under Windows.


roy7:
I added a change to reset() that it will immediately send a connect notification like the Interval does, so we can update our list of games and find games that require moves immediately. The hope is to avoid timeouts if a reset() happens during a game with tight time controls.

roy7:
Issue I had was the process kill in kill() would cause an ECONNRESET uncaught error, which would trigger the shutdown and restart whole script code, but that was causing me to have tons of timeouts.

We do have an on.proc.error set up so I"m not sure why this specific error isn't caught more cleanly by the Bot code itself.

PR #44, opened 2017-05-07
Queue correspondence games to process once at a time

roy7:
I"ll let RoyalLeela test this for a while before merging it, but the code is here for anyone who wants to try it or review.

### DIFF
--- a/gtp2ogs.js
+++ b/gtp2ogs.js
@@ -684,6 +684,7 @@ class Game {
                 this.bot = null;
 
                 if (this.processing) {
+                    this.processing = false;
                     --moves_processing;
                     if (argv.corrqueue && this.state.time_control.speed == "correspondence") {
                         --corr_moves_processing;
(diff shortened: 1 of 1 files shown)


roy7:
TODO:

Choose the game with least time remaining to make moves on first?

Process moves in FIFO order in case multiple people are making moves quickly? Instead of corr_move_pending being true, could be set to a timestamp and choose the lowest each time.

The issue is with only checking for pending moves every 10 seconds, an active player will get all of the moves and the other games will all wait.

Could try to add a function to the Connection object to 'make a pending move'. Like inside the Interval, but call it each time we play a corr move. So we don't wait up to 10 seconds to start working on the next pending move.

roy7:
There's still an issue the an extra gamedata packet, which kills and restarts the bot, can cause corr_moves_processing to decrement twice. Perhaps change the code to not keep a count, but just set a true and false flag? Then an excess =false assignment hurts nothing. It's possible the spurious decrement is after another game starts processing, so it could be possible for two games to process at once with queue enabled.

A different solution could be to not count/track at all, and iterate over the full list of games every interval, and if none are processing but any are pending, then fire off a pending one.

roy7:
An error from uncaught exception like ECONNRESET will result in corr_moves_processing not being decremented. Need to property catch this error (happens if bot crashes, at least on linux) or go with the true/false toggle instead of a counter.

PR #49, opened 2017-09-13
Updated README.md with latest options. gtp2ogs.js version for a GTP dialect.

TokumotoK:
Hope this README.md version is 'friendlier' than before.

Found some bots not understanding the proper "play white q16" gtp format.  They expect "play W q16" instead.  The attached gtp2ogs.js should not be pulled directly into devel.  I hope roy7 will find time to make it an option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Basic usage
 ```
 gtp2ogs --botid <id> --apikey <apikey> <arguments> -- <bot command> <bot arguments>
 ```
- (The space in front of '<bot command>' is important)
+ The space in front of "<bot command>" is important
 
 Options
 =======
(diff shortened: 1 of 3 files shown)


roy7:
Any chance you could check out devel and make the related changes and then pull request that? The current pull request is substantially different than the current code.

roy7:
I know it might make the code more readable, but note the operator priority means () isn't needed after the && tests since && fires later than < does, just FYI.

roy7:
@TokumotoK Sorry I didn't realize this was so close to up to date and ready for merging. I should have pulled it in before my error and corrqueue branches. If you want to rebase one more time I'll get it merged into devel!

Issue #29, opened 2017-03-19
Make it optional for bots to appear in Vs Computer drop down?


roy7:
As more people ask about setting up bots, just to play around with briefly I think, it might be nice for them not to appear in the Vs Computer list as a toggle the operator can set? Maybe they just want to run a program on OGS but not have it accept challenges from the public. My concern is if someone is testing or just wants to have a bot play a specific game, when they stop the bot any other random public challenges will all time out and provide a bad experience to the unsuspecting human opponents?

TokumotoK:
Please allow me to state my take on this issue.  Accepting challenges from the public is (at least should be) a prerequisite to holding a bot account on OGS production server.  Also, it should be a condition to holding a bot account to keep the bot online more or less all the time.  I feel the mods have been too lenient on these points in creating bot accounts.

If someone wants to test a bot, it should be done on the test server, not the production.

anoek:
Implemented in #63

PR #63, opened 2018-03-26
Add --hidden flag to not list the bot in the main computer drop down

anoek:
Implements #29

This capability is currently available on the beta server, I'll merge once it's available in production.

This flag can be set and and cleared whenever the bot operator wants by sending `socket.emit('bot/hidden', true);` or `socket.emit('bot/hidden', false);`

So could in theory be used to hide / unhide the bot from the main challenge list depending on whatever criteria the operator wants (such as unlisting the bot when there's a game in progress). However that is left as an exercise to the reader, this patch just provide a coarse command line option.


roy7:
I know a certain @TokumotoK that might love this.

Issue #58, opened 2018-01-09
Don't connect to finished games


roy7:
See: https://github.com/online-go/gtp2ogs/issues/55#issuecomment-356322452

Since we connect to every active_game, even finished ones, we end up connecting and disconnect()'ing over and over until server stops telling us about that game, which makes the farewell code send a farewell every time we disconnect().

If we check for game status when active_game comes in, we could simply not connect to a finished game in the first place. Then the disconnect() later wouldn't be needed since it wouldn't be in the connected_games list.

windo:
I think this is already the case, should close?

roy7:
If you've never noticed this happening go ahead and close. I just thought sometimes I'd see in my logs old active_games arrive on finished games but we'd connect/disconnect to it even though it's already finished.

wonderingabout:
is this related to #159 ?

wonderingabout:
can this be closed now that #159 is solved ?

roy7:
Sure. We can reopen if it happens again.

PR #72, opened 2018-04-15
Various fixes / additions

lemonsqueeze:
Various fixes / additions i have in my tree, pick and choose what you like:

- --maxsimulgames: Limit maximum number of simultaneous games per player.
  Prevents people from spamming bot by creating 10 games at once =)
- --minmaintime: fischer time support (should probably fix maxmaintime as well)
- Bot.kill() sends 'quit' instead of killing bot, without ill sides effects this time (testing ...)
- Game over handler, if you need to send special command to bot on game-over.
- Split / tidy up onchallenge checks, return one error at a time.

Some parts of gtp2ogs could use some love and cleanup (hint, hint).


roy7:
fischer is already supported by minmaintime. The test you added is the same as the final else test. I didn't add an extra if block for it since the test/result are the same?

roy7:
If gameover() will only be executed one time ever, maybe the farewell message should go in there? Right now it is in disconnect() but old completed games sometimes send out status updates which cause us to reconnect/disconnect, and thus print the farewell multiple times.

That led me to this point for future investigation. https://github.com/online-go/gtp2ogs/issues/58

lemonsqueeze:
> fischer is already supported by minmaintime. The test you added is the same as the final else test. I didn't add an extra if block for it since the test/result are the same?

They're not the same, the fischer part checks `time_control.max_time`, the final else test checks `time_control.main_time`.

Just realized, I put max_time in but for fischer initial time is a better fit for minmaintime really ...
The problem i'm trying to solve is that --minperiods does nothing for fischer time. So bot running with --minperiodtime 30 --minperiods 3 will happily accept fischer 30s max time and timeout if once in a game it takes more than 30s to move. Maybe should have --minperiods check for (fischer max_time >= minperiods * minperiodtime), so 3*30 instead. Or add a --minfischertime ...

lemonsqueeze:
> If gameover() will only be executed one time ever, maybe the farewell message should go in there? Right now it is in disconnect() but old completed games sometimes send out status updates which cause us to reconnect/disconnect, and thus print the farewell multiple times.

Should work, yes.

roy7:
Yes good idea make that leave message contingent on the bot existing. In Fisher time, how about just adding a check that if you specify Max periods or period time Fisher becomes an invalid clock type?

I'm just going from memory here, I'm on my phone, but I thought I treated Fisher like byoyomi with a single period equal to the Fisher time increment. That gets the bot to behave the way we expect.

lemonsqueeze:
> I'm just going from memory here, I'm on my phone, but I thought I treated Fisher like byoyomi with a single period equal to the Fisher time increment. That gets the bot to behave the way we expect.

Yeah, that logic is great we should keep it. Looking at it with a fresh head, for what i'm trying to do simplest is to just specify a --minmaintime really. The rest works great as it is, no need to change it. So the only problem is that --minmaintime is broken with fischer right now =)

I updated my branch:
- Fixed --minmaintime and --maxmaintime to support fischer time.
  Right now it checks both initial_time and max_time. Turns out server won't let you pick max_time smaller than initial_time so could check  initial_time only actually.
- More challenge checks cleanup, almost a breeze to read now.
  js rules, any test involving an undefined property always fails, so no need to check that anything is defined at all.
- Hooked gameover handler in both places and check for bot, so should be pretty solid.

Btw, do you get edited comments when you're on your phone ?
Was wondering if you're seeing the final version or the initial post.

roy7:
I think I have to refresh to see the edited comments, not sure. I'm normally on my PC though. :)

lemonsqueeze:
My main motivation for sending 'quit' is that it allows to detect engine crashes, killing bot makes it pretty much impossible to do outside gtp2ogs.

lemonsqueeze:
Ok, looks like the new gameover hanlder is getting there. Was getting confused by the disconnect() code path. Game over and disconnect are two different things, they shouldn't mix. Testing some more to see how it does.

lemonsqueeze:
Normally on game over we receive:
- `/game/<game_id>/gamedata` with game outcome.
- `active_game` event with phase 'finished' at which point we disconnect from game.

I put some debugging code in, turns out [sometimes](https://github.com/online-go/gtp2ogs/pull/72#issuecomment-381460751) gamedata event arrives _after_ active_game, so it's lost since there's no game to handle it (we already disconnected). That doesn't really make sense, with socket.io over websocket order is supposed to be guaranteed ... Or is there a race condition or something on server side where it'd sending them in reverse order somehow ?
@anoek any ideas ?

roy7:
Sorry for slow feedback. I'll try to read through everything in next night or two. Are the changes you wanted to make complete and ready to merge?

PR #88, opened 2019-01-15
support `--komi any`

windo:
Currently --komi can take explicit values or a `null` value as an
argument. This change adds support for `any` value which accepts any
komi. While there, it also adds an alias of `auto` that works the same
way as `null` but is easier to read for operators and players.


wonderingabout:
great job !! @windo @roy7 

come talk with us on discord #bots channel :+1: 

https://discord.gg/HZ23Cp9

PR #94, opened 2019-01-18
add support for custom board sizes + CONTRIBUTING.md

wonderingabout:
Hi !!

i'm planning to train a 25x1 AI, so i made preparations for that on the gtp2ogs side

now it is possible to have gtp2ogs allow custom boardsize using custom width and height, see readme for details

i inspired myself on @windo 's work on auto and any komi here #93 

general look here : 

https://github.com/wonderingabout/gtp2ogs/tree/custom-boardsize-width-height

possible uses are now : 

- `-- boardsize 9,19` same as before to allow only specified square board sizes, with same default still being `9,13,19`
- NEW ! `--boardsize any` to allow any boardsize
- NEW ! `--boardsize custom --boardsizewidth 25 --boardsizeheight 1,2,3` to allow custom board sizes 25x1 25x2 25x3 (it is possible to allow square board sizes like 19x19 too with this

i also used this opportunity to make clearer the old boardsize messages

has been tested to work on beta ogs successfully, as intended, as you can see below (too tired to comment every screenshot, i hope it's clear)

example of behaviour : 

boardsize z
boardsize y
boardsize x
boardsize w
boardsize v

i can start to train my 25x1 AI now !

last change is i created a contributing.md file (it was too long to fit on the main readme)

feel free to test it everyone and tell me if it's good to you :+1: 

thanks again for all your patience and effort everyone !

thank you for bearing with my struggles and persistent failures and request :+1:


wonderingabout:
also renamed alias `--id` and `--botid` to `--name` and `--botname`

the first time i used gtp2ogs, i thought that botid or id would be the number on ogs profile, for example for this bot : https://online-go.com/player/58441/

username : GnuGo
id : 58441

roy7:
Only change I think is please leave in the old --id and --botid aliases. Adding new alias is fine, but someone might be running a bot using the old alias and we don't want to introduce changes that will break people's existing setups.

wonderingabout:
i understand that it can break old users, but if thats the case i hope they'll notice it quite soon since the bot will fail to start a game

however, the `id` thing can be really confusing especially when gtp2ogs connects "your bot is --number--"

i can still delete it, but if its for older users i think this is not a good investment

wonderingabout:
i find it more important to look at the future

for example, `--komi null` is now `--komi auto` so it will break old bots too

but we still went for it because it is better to use auto in the future

what do you think about it ?

wonderingabout:
also i notice that `id` aliases are not mentioned in the github readme !!

i will add it because it is a possible command to use

roy7:
Also in terms of --komi changes this is why we have a devel branch for testing. The release branches are for solid stuff that won't change. I'll probably want to give all of our recent changes a month or two to be sure no weird bugs appear before publishing a new release, but --id has been in all the prior releases.

wonderingabout:
i understand that it can be confusing to add more data to readme, but if they are confusing, why not just remove the aliases and keep only username ?

like that, if they try botid it will fail and they'll notice immediately

also, most bot admins wont upgrade to devel anytime soon i think

in the end you have the final word @roy7 , but i'm not satisfied how we leave an opportunity to clear up a confusion

hiding the confusion on the readme is not clearing it

roy7:
All right we'll remove the old misleading aliases and if someone complains I'll refer them to you. ;)

wonderingabout:
looooooool

let me commit that :)

i hope bot admins are not as lazy that they'll complain about that having to write 6 more letters

let me clean my branch :)

wonderingabout:
it will be ready in 30 minutes- 1 hour :+1:

wonderingabout:
ready again @roy7 :+1: 

i hope everyone will appreciate the clearer messages i wrote :)

wonderingabout:
did my last commit make it in time lol

wonderingabout:
this is needed to unbreak the readme lol @roy7 

https://github.com/wonderingabout/gtp2ogs/commit/d69d3c84fde97f143ad2e9a68cb9d945325231dd

roy7:
Can you add another PR if something is missing?

wonderingabout:
ok :+1:

Issue #80, opened 2018-12-27
Chinese handicap : "waiting for you to play all your handicap stones" message


wonderingabout:
@roy7 @ anoek 
i noticed several players dont realize they have to play all the handicap stones when the game starts

after they play one stone, they wait a long time, and sometimes they even leave the game

it would be good to add a bot message during the game telling the player that more stones need to be added

for `n` handicap stones, at every `x` number of handicap stone played

```
"you put x handicap stones, please put n-x more handicap stones"
```
example for a 6 handicap stones game: 

```
"you put 1 handicap stones, please put 5 more handicap stones"
```

something based on if and else :)

my most recent example, this player https://online-go.com/game/15885917 waited like 3 minutes before playing his 2nd handicap stone

roy7:
The bot won't even start though until it is the bot's move. OGS indicates to the other player it is their turn. It'll show they have a game waiting in top right corner. I don't think this is a gtp2ogs issue.

wonderingabout:
i noticed this behaviour in several games though

the problem is since the clock doesnt start and nothing moves, the human player sometimes assume the bot is thinking long for its first move, or the bot has crashed and leaves

my first message may have been complicated, but i thought of something easier : 

if 
rules = chinese and notification.handicap > 0
write bot greeting message
"hello, this is a handicap game, please put your notification.handicap stones on the board"

see screenshots below

![handi1](https://user-images.githubusercontent.com/38690718/50508166-fbe9b580-0a80-11e9-94d1-b8ec184cb719.png)
![handi2](https://user-images.githubusercontent.com/38690718/50508165-fbe9b580-0a80-11e9-83ff-0c5c703b254a.png)
![handi3](https://user-images.githubusercontent.com/38690718/50508164-fbe9b580-0a80-11e9-8bf2-ce46ff43414c.png)
![handi4](https://user-images.githubusercontent.com/38690718/50508163-fbe9b580-0a80-11e9-840b-6949b76e5a8f.png)

it definitely can be confusing from a human player perspective

roy7:
Ahh ok I thought clock started when you placed the first stone, my mistake.

wonderingabout:
its ok :) 
(mail notifications with thunderbird are sure helpful)

wonderingabout:
happened again today...

i'll try to implement this on gtp2ogs

wonderingabout:
fixed in #118 

example 

[...]

PR #96, opened 2019-01-20
place_free_handicap

### LINKED ISSUE #70: place_free_handicap
Follow up on #65 for the case where bot is playing black with handicap:

So we need to send place_free_handicap instead of genmove at the beginning. It returns all the stones at once so probably have to store them somewhere and return them one by one to server when asked. I'm not super familiar with gtp2ogs code so first idea was to re-engineer makeMove() to do it behind the scene and keep current interface, this way the rest of the code doesn't have to change.

lemonsqueeze:
Hi,

Just saw there was another PR for place_free_handicap so i though i might join in =)

This is a rebase of an earlier experiment with place_free_handicap:
The idea was to decouple `makeMove()` into `getBotMoves()` + `uploadMove()`.
Implementing place_free_handicap becomes easy once this is done.

Haven't tested this rebase but the old branch was working well iirc.

Fixes #70 .

### DIFF
--- a/gtp2ogs.js
+++ b/gtp2ogs.js
@@ -659,9 +659,12 @@ class Bot {
             if (eb) eb(e);
         }
     } /* }}} */
+
+    // For commands like genmove, place_free_handicap ... :
+    // Send @cmd to engine and call @cb with returned moves.
     // TODO: We may want to have a timeout here, in case bot crashes. Set it before this.command, clear it in the callback?
     //
-    genmove(state, cb) { /* {{{ */
+    getMoves(cmd, state, cb, eb) { /* {{{ */
         // Do this here so we only do it once, plus if there is a long delay between clock message and move message, we'll
         // subtract that missing time from what we tell the bot.
         //
@@ -672,9 +675,14 @@ class Bot {
         //
         this.firstmove = false;
 
-        this.command("genmove " + this.game.my_color,
-            (move) => {
-                move = typeof(move) == "string" ? move.toLowerCase() : null;
+	this.command(cmd, (line) => {
+	    line = typeof(line) == "string" ? line.toLowerCase() : null;
+	    let parts = line.split(/ +/);
+	    let moves = [];
+	    
+	    for (let i=0; i < parts.length; i++) {
+		let move = parts[i];
+		
                 let resign = move == 'resign';
                 let pass = move == 'pass';
                 let x=-1, y=-1;
@@ -683,16 +691,20 @@ class Bot {
                         x = gtpchar2num(move[0]);
                         y = state.width - parseInt(move.substr(1))
                     } else {
-                        this.log("genmove failed, resigning");
+                        this.log(cmd + " failed, resigning");
                         resign = true;
                     }
                 }
-                cb({'x': x, 'y': y, 'text': move, 'resign': resign, 'pass': pass});
-            },
-            null,
+                moves.push({'x': x, 'y': y, 'text': move, 'resign': resign, 'pass': pass});
+	    }
+
+	    cb(moves);
+        },
+            eb,
             true /* final command */
         )
     } /* }}} */
+
     kill() { /* {{{ */
         this.log("Stopping bot ");
     this.ignore = true;  // Prevent race conditions / inconsistencies. Could be in the middle of genmove ...
@@ -913,42 +925,8 @@ class Game {
             'game_id': game_id
         }));
     } /* }}} */
-    makeMove(move_number) { /* {{{ */
-        if (DEBUG && this.state) { this.log("makeMove", move_number, "is", this.state.moves.length, "!=", move_number, "?"); }
-        if (!this.state || this.state.moves.length != move_number) {
-            return;
-        }
-        if (this.state.phase != 'play') {
-            return;
-        }
 
-        ++moves_processing;
-        this.processing = true;
-        if (argv.corrqueue && this.state.time_control.speed == "correspondence") {
-            ++corr_moves_processing;
-        }
-
-        let passed = false;
-        let passAndRestart = () => {
-            if (!passed) {
-                passed = true;
-                this.log("Bot process crashed, state was");
-                this.log(this.state);
-                this.socket.emit('game/move', this.auth({
-                    'game_id': this.state.game_id,
-                    'move': ".."
-                }));
-                this.procesing = false;
-                --moves_processing;
-                if (argv.corrqueue && this.state.time_control.speed == "correspondence") {
-                   this.corr_move_pending = false;
-                    --corr_moves_processing;
-                }
-                if (this.bot) this.bot.kill();
-                this.bot = null;
-            }
-        }
-
+    startBot(eb) { /* {{{ */
         if (!this.bot) {
             this.bot = new Bot(this.conn, this, bot_command);
             this.log("Starting new bot process [" + this.bot.proc.pid + "]");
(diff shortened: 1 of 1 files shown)


lemonsqueeze:
As far as i know #80 is not a gtp2ogs issue, it happens in human games played with chinese rules also. Try creating a [server](https://github.com/online-go/online-go.com) issue, maybe anoek can do something about it.

wonderingabout:
it's true that it can also happen on human vs human games, but if thats the case the humans will messge each other quite quickly

however a bot will patiently and silently sit until condition is fulfilled, resulting in the situations we see in human vs bot handicap games H2+

while adding a stone message notification on human vs human is feasable, i think it's not at all required

however, again, on bot vs bot, the situation is slightly different

gtp2ogs.js
```diff
@@ -1017,10 +1021,36 @@ class Game {
...
+		if (this.bot) this.bot.kill();
+		this.bot = null;
+		this.uploadMove({'resign': true});
+		return;
+	    }
+	    this.handicap_moves = moves;
+	    this.uploadMove(this.handicap_moves.shift());
+	};
+	    
+	this.getBotMoves("place_free_handicap " + this.state.handicap, storeMoves, sendPass);
```
windo: You should probably add a command line flag to request the use of `place_free_handicap`, I think @wonderingabout was worried that phoenixgo might not implement that GTP command?

gtp2ogs.js
```diff
@@ -672,9 +675,14 @@ class Bot {
         //
         this.firstmove = false;
 
-        this.command("genmove " + this.game.my_color,
-            (move) => {
-                move = typeof(move) == "string" ? move.toLowerCase() : null;
+	this.command(cmd, (line) => {
+	    line = typeof(line) == "string" ? line.toLowerCase() : null;
```
windo: I know it was this way before, but should the second option for the tenary operator be `""` rather than `null`? Otherwise the `split` below is going to throw an exception.

gtp2ogs.js
```diff
@@ -958,44 +936,94 @@ class Game {
...
+    //
+    getBotMoves(cmd, cb, eb) { /* {{{ */
+        ++moves_processing;
+        this.processing = true;
+        if (argv.corrqueue && this.state.time_control.speed == "correspondence")
+            ++corr_moves_processing;
 
-        this.bot.genmove(this.state, (move) => {
-            this.processing = false;
+	let doneProcessing = () => {
```
windo: Nice cleanup, thanks for capturing this in a function!

gtp2ogs.js
```diff
@@ -958,44 +936,94 @@ class Game {
...
+            this.bot = null;
+	    if (eb) eb(e);
         }
-            }
-            if (!PERSIST && this.bot != null) {
+
+        if (!this.bot)  this.startBot(botError);
+
+	if (DEBUG) this.bot.log("Generating move for game", this.game_id);
+	this.log("genmove " + (this.my_color == "black" ? "b" : "w"));
```
windo: This log message should be updated to use cmd too.

gtp2ogs.js
```diff
@@ -1017,10 +1021,36 @@ class Game {
...
+	// Get handicap stones from bot and return first one.
+	let storeMoves = (moves) => {
+	    if (moves.length != this.state.handicap) {  // Sanity check
+		this.log("place_free_handicap returned wrong number of handicap stones, resigning.");
+		if (this.bot) this.bot.kill();
+		this.bot = null;
+		this.uploadMove({'resign': true});
+		return;
+	    }
+	    this.handicap_moves = moves;
```
windo: The bot I'm working on used to generate a PASS as a handicap move from time to time - if given way too many handicap moves to play. In that case, `gtp2ogs` would just hang as the server will ignore the pass as a handicap.

I think with this commit, that also happens because we upload the pass and the server will (correctly) reject it.

Arguably, bots should implement the spec correctly, but it would be helpful to bot authors if `gtp2ogs` would check that there are not `pass` moves and complain and resign the game if there are any. Large handicaps come up rarely so it is relatively easier to miss that case during development.

gtp2ogs.js
```diff
@@ -958,44 +936,94 @@ class Game {
...
         }
-            }
-            if (!PERSIST && this.bot != null) {
+
+        if (!this.bot)  this.startBot(botError);
+
+	if (DEBUG) this.bot.log("Generating move for game", this.game_id);
+	this.log("genmove " + (this.my_color == "black" ? "b" : "w"));
+
+        this.bot.getMoves(cmd, this.state, get_handicap_stones, (moves) => {
```
windo: What is `get_handicap_stones` here, I assume this comes up in the next commit? The signature of `getMoves` above has 4 argument, this call has 5?

gtp2ogs.js
```diff
@@ -958,44 +936,94 @@ class Game {
...
+    // Get move from bot and upload to server.
+    //
+    makeMove(move_number) { /* {{{ */
+        if (DEBUG && this.state) { this.log("makeMove", move_number, "is", this.state.moves.length, "!=", move_number, "?"); }
+        if (!this.state || this.state.moves.length != move_number)
+            return;
+        if (this.state.phase != 'play')
+            return;
+
+	let sendMove = (moves) => {  this.uploadMove(moves[0]);  };
```
windo: Should this be called sendFirstMove as it silently ignores the rest of the moves?

gtp2ogs.js
```diff
@@ -1017,10 +1021,36 @@ class Game {
...
-	this.getBotMoves("genmove " + this.my_color, sendMove, sendPass);
+	if (!doing_handicap) {  // Regular genmove ...
+	    let sendMove = (moves) => {  this.uploadMove(moves[0]);  };
+	    this.getBotMoves("genmove " + this.my_color, sendMove, sendPass);
+	    return;
+	}
+	
+	// Already have handicap stones ? Return next one.
+	if (this.handicap_moves.length) {
+	    this.uploadMove(this.handicap_moves.shift());
```
windo: What is the advantage of storing and sending them one at a time over sending them all straight away? The latter also seemed to work and would allow us to skip the move storage.

windo:
Oh, I briefly tested this branch with with my test bot and can confirm it correctly handles handicap stones on beta ogs.

wonderingabout:
@windo 

currently, the current devel branch supports handicap in a satisfying way (i tested a week ago against gnugo as black or white, and stones are placed freely by both bots as black or white, the issue is with komi 7.5 required)

as for the GTP command, i'm indeed not sure if phoenixgo supports it, i guess i need to run `list_commands` right ?

wonderingabout:
here are the supported commands : 

supported commands

gtp2ogs.js
lemonsqueeze: Then phoenixgo is broken. Please don't use gtp2ogs to work around bugs in bots ! If a bot admin wants to add such code to his copy of gtp2ogs he is free to do so, but i wouldn't want that in the repository. At any rate gtp2ogs should implement the gtp spec correctly by default (use place_free_handicap).

gtp2ogs.js
lemonsqueeze: Ah, i knew something got clobbered in the rebase =)

gtp2ogs.js
lemonsqueeze: I'd say same, don't use gtp2ogs to fix bugs in your bot. Fix your bot instead ! Or do some workarounds but keep it a secret =)

gtp2ogs.js
lemonsqueeze: Yes, it disappears in the next commit. Rebase madness, i'll remove it there too.

gtp2ogs.js
lemonsqueeze: Or sendTheMove(), anything is good with me.

gtp2ogs.js
lemonsqueeze: Oh ok, let's send them all together then. I guess i was trying to keep it the same from server side point of view.

gtp2ogs.js
lemonsqueeze: I'd do this in another PR / commit then. One change at a time ...

gtp2ogs.js
wonderingabout: phoenixgo supports free handicap as is currently on devel branch, as black and as white without any issue (to give or take handi)

the only thing i pointed is the behaviour of human waiting long or leaving games waiting for bot to play while they didnt put all handicap stones, as explained in #80 

you are right that we shouldnt hack gtp2ogs, custom branches can achieve that

gtp2ogs.js
wonderingabout: as said in other comment :+1: 

phoenixgo supports free handicap as is currently on devel branch, as black and as white without any issue (to give or take handi)

so phoenixgo does not need this issue, it works fine already
also, phoenix go is unlikely to take handicap from anyone on ogs (also its bad at handi to begin with)

the only thing i pointed is the behaviour of human waiting long or leaving games waiting for bot to play while they didnt put all handicap stones, as explained in #80 

you are right that we shouldnt hack gtp2ogs, custom branches can achieve that

thanks for your effort everyone :+1:

wonderingabout:
as added on other comments in the review : 

phoenixgo supports free handicap as is currently on devel branch, as black and as white without any issue (to give or take handi)

so phoenixgo does not need this issue, it works fine already
also, phoenix go is unlikely to take handicap from anyone on ogs (also its bad at handi to begin with)

the only thing i pointed is the behaviour of human waiting long or leaving games waiting for bot to play while they didnt put all handicap stones, as explained in #80 

you are right that we shouldnt hack gtp2ogs, custom branches can achieve that

thanks for your effort everyone :+1:

lemonsqueeze:
@windo 
Thanks for taking the time to review this.

I pushed the new branch with changes above except sending moves together.
Sorry i don't have much time to dedicate to gtp2ogs at the moment, feel free to hack on top of it if you like.

Good to see place_free_handicap being addressed at last !

gtp2ogs.js
windo: Well, you check above that the correct number of handicap stones is returned by the bot, to avoid getting into the state where you send a partial handicap to the server and then get stuck forever (if I understand the code right).

Sending one or more `pass` moves as part of handicap is the same as sending a shorter handicap and `gtp2ogs` will get stuck. The user will wait for the bot to move and nothing will happen. The game will time out and the player will be annoyed.

So we should just validate the input from the bot here and fail fast if the input was incorrect. The spec says:

https://www.lysator.liu.se/~gunnar/gtp/gtp2-spec-draft2/gtp2-spec.html#SECTION00073200000000000000

```
Vertices must not be repeated or include ``pass''.
```

windo:
FWIW I've been running this branch on my bots for the last few days and have not observed it failing. Though there are not many games being played with handicap stones (at least public ones), at least this very short game came up:

https://online-go.com/game/16222392

gtp2ogs.js
lemonsqueeze: Yes, we should do this. Sorry for misunderstanding your previous comment, this is exposing bugs in bots, not hiding them.

lemonsqueeze:
Added code to check for pass.

Sending all the moves together doesn't seem to work, at least not like [this](https://github.com/lemonsqueeze/gtp2ogs/tree/place_free_handicap_bad). gtp2ogs gets confused and bot ends up with multiple place_free_handicap commands...

windo:
Even if you send them all, the server will send all of the moves back individually too. So you'd have to ignore moves received from the server until they have caught up with the handicap.

But I don't want to get in the way of this PR getting in - storing and sending one-by-one is fine too, maybe we should get this in first (as it works right now) and then if someone (say - me) wants to send them all at once that can be another PR.

roy7:
Ok so @windo and @lemonsqueeze both think this is good to merge as-is? Looking at the earlier conversations, did we not need to add the change to use "" instead of null to avoid an error on split( ) later?

Oh wait it says "Outdated" now so I guess it was already fixed. :)

gtp2ogs.js
roy7: Oh I see your point, the old code was already using null. So leave that behavior as-is and we can PR it separately than the rest of these changes, correct?

windo:
I believe that issue still exists. But it also existed before this PR. And honestly, I'm unsure what that entire line is supposed to accomplish, why would the result from the command be anything but a string?

roy7:
I'm not sure, I think maybe that code has been around a long time. Perhaps it was just a sanity/error check. It can be cleaned up if someone wants. :) I'll go ahead and merge the rest of this now. Thank you for the work and testing! There was a long period of time no one worked on gtp2ogs except me so it's a relief to have some other contributors active. :)

wonderingabout:
> There was a long period of time no one worked on gtp2ogs except me so it's a relief to have some other contributors active. :)

whaaaaaaaat
i'm so jealous looool

how about this @roy7 :rocket: https://github.com/online-go/gtp2ogs/pull/94

PR #99, opened 2019-01-26
Split up gtp2ogs.js and add a basic smoke test

### LINKED ISSUE #97: No tests for gtp2ogs
It can be tricky to validate that changes don't cause regressions without any test coverage. At the very least, I think there should be some "smoke tests" for the integration of all the parts.

I'd like to split the `gtp2ogs.js` file into a "main file" and library file(s) (say `bot.js`, `game.js` and `connection.js`) so that the code could be loaded into test cases without triggering all of it to run end-to-end immediately. Then I'd like to add some basic test coverage that all the parts fit together - a basic integration test that mocks the GTP program and OGS server to behave well and confirm that `gtp2ogs` behaves relaying between the two. 

I haven't done a whole lot of node.js development, is `mocha` an acceptable module to use for testing? And `simple-mock` for mocking the GTP process and OGS server?

Does splitting up the main file into a few separately loadable files sound OK? Or should I try to keep as much of it together as possible and try to separate a minimal "entry point" file?

windo:
If there are better ideas on how to do the split, please do suggest.

Fixes #97.


wonderingabout:
we'll need to wait from advice and testing from knowledgeable people, as well as people who worked on gtp2ogs code so far :+1: 

@roy7 @anoek @lemonsqueeze @Dorus @thouis @thouis

windo:
@wonderingabout, I'm not very familiar with the documentation - what do you think should be updated?

IDK if it is the right time to set up automatic testing or anything at this point, but maybe should suggest an `npm test` run locally as part of development to make sure the changes haven't broken anything basic?

I don't know how the setup instructions need to change - I had been running it locally with `node gtp2ogs --arguments --here` but I think the official suggested way is to `npm install` it, which might now work differently with multiple files.

wonderingabout:
as far as documentation is concerned, you can let me handle it :+1: 

so far i would need to update the tutorial first thing
then, there shouldnt be much to change for documentation actually

basically we're only splittting the main file into several modules, so we'd need to add small explanations about how to handle that for users and contributors

windo:
(rebased and fixed a few errors that had slipped in during refactoring)

roy7:
Ok rebase a last time and we'll get this merged later this weekend. I'll probably move this current devel into master before hand.

No more PRs from @wonderingabout or @Dorus until after this is in so it doesn't keep delaying you. ;)

wonderingabout:
yes i can do it afterwards, but it will take me only 1 day, and it would be great if its in the next master release

so i would want to avoid @windo the hassle to rebase again after i commit :1st_place_medal:

windo:
I've rebased again. I ran `meld` and compared the original gtp2ogs.js file to the split up files and couldn't spot anything that I had missed in the split.

wonderingabout:
done here #111

roy7:
I'll merge this tomorrow. Since the split alone shouldn't really break anything we can prob just keep everything in devel a while longer. That will let @wonderingabout keep adding stuff before the next master release as well.

roy7:
Sorry for the delay. Here we go!

windo:
To make the build status little icon thing work, someone with permissions on the project needs to go to travis-ci.org and activate it for the project.

roy7:
@anoek I might have permission but I have no idea how to do that, might you do it for us? :)

roy7:
I tried logging into travis-ci.org but I don't think I'm able to mess with that setting myself.

PR #101, opened 2019-01-26
add a --logfile option

windo:
This makes it more convenient to track what your bot was doing for reports of issues that happened some time ago.

### DIFF
--- a/gtp2ogs.js
+++ b/gtp2ogs.js
@@ -39,6 +39,7 @@ let optimist = require("optimist")
     .describe('insecure', "Don't use ssl to connect to the ggs/rest servers")
     .describe('beta', 'Connect to the beta server (sets ggs/rest hosts to the beta server)')
     .describe('debug', 'Output GTP command and responses from your Go engine')
+    .describe('logfile', 'In addition to logging to the console, also log to this file')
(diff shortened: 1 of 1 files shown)


wonderingabout:
this is so amazing really @windo 

yes we should have added this a lot sooner

i suggest renaming it `--logtofile` similar to what phoenixgo uses `--logtostderr`

wonderingabout:
ok @roy7 

btw this gave me the idea of adding support to run several bots with just on gtp2ogs command

my purpose for doing this is for example if we replay an AI tournament, or even just for live games, i want my GPU to play 1 move at a time, yet spread the moves evely between bots

for example

bot 1 
- game a
- game b
- game c
bot 2 
- game z
- game y

i want gtp2ogs to handle my gpu power if "one game at a time is specified"

wonderingabout:
also @roy7 

i'm already adding PR to other projects lol
i think its faster if you do it from the website :) (not need to clone and all, since you have write access to the project)

wonderingabout:
actually @roy7 i'll handle the readme next time :+1: 

there are 2-3 other things i want to add too, so it's a good opportnity
- fix notes link
- add a short intro about what being a bot admin is (total control, online when you want, restrict power, etc..), which is vague to new users

PR #105, opened 2019-01-31
Add tests for a couple of conditions and fix the issues so the tests pass

windo:
This queue currently fixes 3 ~problems (and it's intended to be pulled on top of #99):

1. Pick a random game from correspondence queue rather than always the first one - making the the queue fairer in case multiple correspondence games are in flight at the same time
2. Never timeout a connection when it is the bot's turn to play, removing a condition under which a game can be left hanging and timing out. Also simplifies timeout handling.
3. Call the error callback from bot commands if the bot exits (crashes) so Game could restart the bot and try again.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -86,15 +86,19 @@ class Connection {
                 if (Game.corr_moves_processing == 0) {
                     // Choose a corr game to make a move
                     // TODO: Choose the game with least time remaining
-                    //
+                    let candidates = [];
                     for (let game_id in this.connected_games) {
                         if (this.connected_games[game_id].corr_move_pending) {
-                            this.connected_games[game_id].makeMove(this.connected_games[game_id].state.moves.length);
-                            break;
+                            candidates.push(this.connected_games[game_id]);
                         }
                     }
+                    // Pick a random game that needs a move.
+                    if (candidates.length > 0) {
+                        let game = candidates[Math.floor(Math.random()*candidates.length)];
+                        game.makeMove(game.state.moves.length);
+                    }
                 }
-            }, 10000);
+            }, 1000);
         }
 
         this.notification_connect_interval = setInterval(() => {
--- a/connection.js
+++ b/connection.js
@@ -37,7 +37,6 @@ class Connection {
         });
 
         this.connected_games = {};
-        this.connected_game_timeouts = {};
         this.games_by_player = {};     // Keep track of active games per player
         this.connected = false;
 
@@ -48,6 +47,10 @@ class Connection {
             }
         }, (/online-go.com$/.test(config.host)) ? 5000 : 500);
 
+        if (config.timeout) {
+            this.idle_game_timeout_interval = setInterval(
+                this.disconnectIdleGames.bind(this), 1000);
+        }
 
         this.clock_drift = 0;
         this.network_latency = 0;
@@ -118,13 +121,8 @@ class Connection {
             this.connected = false;
 
             conn_log("Disconnected from server");
-            if (config.timeout)
-            {
-                for (let game_id in this.connected_game_timeouts)
-                {
-                    if (config.DEBUG) conn_log("clearTimeout because disconnect from server", game_id);
-                    clearTimeout(this.connected_game_timeouts[game_id]);
-                }
+            if (config.timeout) {
+                clearInterval(this.idle_game_timeout_interval);
             }
 
             for (let game_id in this.connected_games) {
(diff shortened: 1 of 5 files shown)


roy7:
The old corrqueue behavior was unpredictable and probably doesn't need to be kept even as an option. I wonder though if games waiting in corrqueue should go into an actual queue and then if a game is waiting, we pop it off the front? So it's always FIFO instead of random or arbitrary (current way).

wonderingabout:
i'm happy with the new one, but i remember that @nerai (forgot his github name) appreciated some of the features of the first corrqueue, maybe ask him :+1: 

also, as you said random is better than chaotic, but personally i would prefer a "dynamic circle" order : 

some code i just came by in my head to illustrate :+1: 

games A,B,C,D,E

```
while (number of connected games >=1 && argv.corrqueue) {
    sort games by some number order; // A=1 then B=2 then C=3 then D=4 then E=5, n times
    let gameToPlay = "1"; // we start with the first game in the list
    if (turn to play in gameToPlay) {
        if (gameToPlay opponent doesnt play in less than 30 seconds) {
            skip gameToPlay;
        }
    } else { 
        play gameToPlay;
    gameToPlay = (gameToPlay%n)+1;
    }
}
// like that, after the queue is in circle, but its also dynamic :)
```

nerai:
Hi, you found the right github handle, so I noticed this thread.

Regarding my opinion about the corrqueue, I found that the behavior in which it plays in the order of most recent (i.e. staying in the games as long as possible) is not so good. It would be more sensible to play in e.g.

* the *least* recent games first,
* or to play the the game which has the smallest time for the move remaining,
* or to select randomly as this PR apparently does.

wonderingabout:
personally i would also want the least recent games first, random is better than pseudo-live but doesnt suit my needs the most (corr AI tournament with equally spread game moves on every game)

but i guess depending on the bots and usages one may want some partially permanent presence

windo:
I don't have a particular reason to be fond of random choice - it just sounded good enough for my purposes. I might take a quick look to see if it's super easy to introduce a proper queue, but if not, then I'd rather get this merged and we can look into the queue implementation afterwards.

wonderingabout:
sure, i understand, do it in an efficient and reasonable way :+1:

roy7:
#99 is merged so I assume this branch needs updated or rebased in some way, the changed file diffs are really large.

windo:
I have been running this branch for a while and haven't seen any more crashes. I've still seen some cases where the bot has gotten stuck, so I'm ambivalent on whether it is worth merging this into devel right now.

I think the first two commits should fix some cases that I've observed multiple times as being bad. The bot crash handling commit I think does not hurt but I know it has shortcomings - I had some more handicap bugs in my bot that caused it to crash sometimes for large handicaps and those games still got stuck.

So could merge all three commits and someone (me or someone else, if the code is available for all on devel) can hopefully improve the bot crash handling if they see more crashes. Or I can move the third commit out of the PR and keep working it and send another PR when I think that's ready.

WDYT?

wonderingabout:
@windo 

i dont know what the others think, but this looks wiser to me : 

> Or I can move the third commit out of the PR and keep working it and send another PR when I think that's ready.

Because we can always add the other feature later when it's fixed, no rush on that, but if we implement on devel a buggy feature, then we'll be pincered between fixing it now or focusing on other things

roy7:
Just a general guideline for anyone, use 4 space indent without tabs please. Just to keep it all unified.

windo:
Fixed indentation and moved the bot crash handling off the PR.

Please take a look!

connection.js
```diff
@@ -234,24 +198,25 @@ class Connection {
...
+    disconnectIdleGames() {
+        let now = Date.now();
+        for (let game_id in this.connected_games) {
+            let state = this.connected_games[game_id].state;
+            if (state == null) {
+                if (config.DEBUG) conn_log("No game state, not checking idle status for", game_id);
+                continue;
+            }
+            if ((state.clock.current_player != this.bot_id) && (state.clock.last_move + config.timeout < now)) {
+                if (config.DEBUG) conn_log("Disconnecting from game, other player has been idling for", config.timeout);
```
roy7: Do we need a space after 'for' here at the end?

roy7:
Looked it all over this morning, looks good. I put in one comment for a possible missing space in a log message.

I like the idea of just checking the list every so often (once per second right?) instead of trying to manage setting/clearing Timeouts constantly. The Timeouts mean we don't waste cpu time, but cpu time is cheap these days. And it avoids the risk of a stray Timeout running around.

Should we also handle pauses in this way? Instead of setting pause timeouts when we pause and clearing when they unpause, just iterate and check for pause maximums exceeded?

game.js
```diff
@@ -150,6 +150,14 @@ class Game {
         this.socket.on('game/' + game_id + '/move', (move) => {
             if (!this.connected) return;
             if (config.DEBUG) this.log("game/" + game_id + "/move:", move);
+            if (!this.state) {
+                console.error("Received move for " + this.game_id + "but no state exists");
```
Dorus: Need a space before "but"?

roy7:
There's nothing async in game connection/gamedata loading right? So we can't be in a state where a connection comes in and interval fires and we try to makemove on a game that isn't actually initialized yet?

roy7:
I'll merge then fix those 2 whitespace issues.

windo:
I don't have a whole lot of experience with node.js and not sure I understand the threading model 100% correctly. Is it that any event callback will run completely from start to end before any other event callback is triggered? So concurrency issues can only arise if two separate callbacks assume they will be running back-to-back with no other event (in this case, game timeout processing) happening between them?

So what I should check is that for example between getting an `active_game` and subsequent `gamedata` the timeout handling will not do something silly? For that, I should be able to add a test case technically, not sure if it would be annoying to set up.

Will fix the log line spacing issues!

windo:
Oh, you merged already, thanks!

LMK if I there was still something I should clean up!

wonderingabout:
@windo maybe @Dorus has some knowledge to share on nodejs

roy7:
Dorus asked on chat if it was possible for the Interval to fire in the middle of a new game connecting arriving and maybe try to do something on a connected game that doesn't have state set up yet. I'm thinking it's okay though because we make sure game.state exists and I don't believe there are any async actions happening inside the game state setup/initialization.

Since nodejs is single threaded, any non-async sections will fully finish before anything else happens.

Dorus:
@windo Yes you are correct, callbacks will always run to completion and the only time others 'threads' can take over in js is when you are waiting for an async function/callback somewhere.

PR #115, opened 2019-02-04
UHMAEAT : Universal Highly Modulable And Expandable Argv Tree

wonderingabout:
- Tons of benefits over old messages

- about simple timecontrol, since i didnt understand the link between maintime and periodtime, i disabled it
correct me if i did wrong, but here is the problem :
it has a main time with t.main_time
but ??
it also has a period time too with t.per_move ??

so it has both a main and a period time, but no period time ??
how bot admins handle it ?
does it depend on maintime or periodtime ??
is it really safe and worth using for bots on ogs ?


roy7:
Ok I'll merge as-is but add an Issue would you about support for simple time control need added?

wonderingabout:
opening an issue atm

wonderingabout:
https://github.com/online-go/gtp2ogs/issues/116

Issue #119, opened 2019-02-05
Main things to prioritirize before next master release


wonderingabout:
So :1234: 

history tells us : 
- release 5.1 : march 2017
- release 5.3 : march 2018
- release x.x : xxx xxxx

also a lot of new features, improvements, and fixes, have been released since that time, too much to list them all here

but maybe some have words of important things we may want to look again at before making a new major release 

mostly directed to code contributors so far, but more generally to everyone
@anoek @roy7 @Dorus @windo @lemonsqueeze @tokumotoK

as for me i want to look again at : 

- autohandicap issue in #28 : using `--noautohandicap` removes the issue but doesnt solve it, we want to make automatic handicap dynamically detect min and max handicap, and adjust to rank difference
- simple time support in #116 : does it use main time `t.main_time` or period time `t.per_move`, and how to support it again on gtp2ogs ?
- challenge 404 error #100 and #52 : does not crash the bot so far so been given a low priority, but still occuring every while, sometimes it even loops when gtp2ogs fails to reject a game for some reason, we may want to look again at log behaviour as well as reject behaviour
- remove most warnings and replace them with defaults
- rename `--maxtotalgames` and `--maxactivegames` to the clearer `connected` games
- @roy7 maybe you want to have a look again at the `--persist` behaviour for corr games ? in https://github.com/online-go/gtp2ogs/issues/59
- using a config file instead of argv command line, is easily exportable and much more flexible to add new options, but also easier to read and edit, in #125 

will be updated as we add or point more things

feel free to use this issue as a more general discussion of what we want to do more generally before the release and after

TokumotoK:
#56 --maxpausetime is huge in my mind. 

It appears I have failed to get the point across to roy7 and wonderingabout how important it is to allow (strong) human players to pause a game against bots without forcing a long time setting.  I feel I faced a flat rejection to understand the issue, but am willing to help understanding further. 

Even a simple 24 hour clock for --limitpause without any time setting option is a big improvement, and is a shame not to be included in the next big release.

roy7:
I guess I was thinking if you wanted to let people pause then there was no harm in letting them pause indefinitely? But yeah adding a timeout for max pause length seems like a good option to add.

TokumotoK:
Perhaps viewing GnuGo or Fuego user profile page would convince you how so many (mostly weak) users pause a game and forget.  This makes loading the bot profile page a hassle, and we do depend on the profile page to be viewed by bot users for limitations.

[...]

PR #120, opened 2019-02-06
(unfinished) initial attempt at maxpauses and maxpausetime

wonderingabout:
still unfinished

see #119 

currently getting undefined pauses error, maybe @Dorus or someone wants to review

### DIFF
--- a/game.js
+++ b/game.js
@@ -113,6 +113,42 @@ class Game {
                 this.resumeGame();
             }
 
+            // the code below can be viewed as a maxperiods + maxperiodtime problem, where period=pause
+            // if maxpausetime and 
+            if ((config.maxpausetime || config.maxpauses) && clock.pause.paused && !clock.pause.pause_control["stone-removal"] && !clock.pause.pause_control.system && !clock.pause.pause_control.weekend && !clock.pause.pause_control["vacation-" + clock.black_player_id] && !clock.pause.pause_control["vacation-" + clock.white_player_id]) {
+
+                let pausesNumberConsumed = 0; // max is config.maxpauses
+                let remainingPausesNumber = config.maxpauses - pausesNumberConsumed; // when = 0, no more pauses allowed
+        
+                function pausetimeCountdown() {
+                    if (remainingPausesNumber = 0) {
+                        conn_log(`All ${config.maxpauses} have been consumed, no more pauses allowed, resuming the game`);
+                        this.sendchat(`All ${config.maxpauses} pauses allowed have been consumed, no more pauses allowed, resuming the game`);
+                    } else {
+                        let i = Date.now(); // current time, will be refreshed every 1 minutes in the loop
+                        let pausetimeInitialTime = Date.now(); // date at which we start pausing the game
+                        let pausetimeMaxTime = config.maxpausetime + Date.now(); // after the pause starts, the human player has  config.maxpausetime seconds before the game resumes
+                        // we send an ingame sendchat to notify the user of pause settings allowed, in human readable time
+                        conn_log(`The game has been paused, pausetime allowed : ${timespanToDisplayString(config.maxpausetime)}, remaining pauses allowed : ${timespanToDisplayString(config.maxpauses)}`);
+                        this.sendchat(`The game has been paused, pausetime allowed : ${timespanToDisplayString(config.maxpausetime)}, remaining pauses allowed : ${timespanToDisplayString(config.maxpauses)}`);
+
+                        let pausetimeCheck = function() {
+                            if (i >= pausetimeMaxTime) { // when all allowed pause time is consumed
+                                pausesNumberConsumed+=1;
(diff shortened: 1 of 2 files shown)


roy7:
I've done so little with timeouts that I don't recall if the callback function is going to have access to those variables from out of its scope at the time it runs, or if we need to pass them to the callback. We might also need to store the pause timeout callback like we do other timeouts and clear them once unpaused or if the game is ended (if someone resigns during the pause).

game.js
```diff
@@ -113,6 +113,42 @@ class Game {
                 this.resumeGame();
             }
 
+            // the code below can be viewed as a maxperiods + maxperiodtime problem, where period=pause
+            // if maxpausetime and 
+            if ((config.maxpausetime || config.maxpauses) && clock.pause.paused && !clock.pause.pause_control["stone-removal"] && !clock.pause.pause_control.system && !clock.pause.pause_control.weekend && !clock.pause.pause_control["vacation-" + clock.black_player_id] && !clock.pause.pause_control["vacation-" + clock.white_player_id]) {
+
+                let pausesNumberConsumed = 0; // max is config.maxpauses
```
roy7: Move this up to around line 25 or so. It should be in the initial variables of the constuctor, so that we can store it and access it later as part of this game object. It won't be recognized otherwise down inside the pausetimeCheck() callback, as I understand things. Someone else speak up if I'm wrong.

roy7: You'd do like ```this.pausesNumberConsumed = 0;```

game.js
```diff
@@ -113,6 +113,42 @@ class Game {
...
+                    } else {
+                        let i = Date.now(); // current time, will be refreshed every 1 minutes in the loop
+                        let pausetimeInitialTime = Date.now(); // date at which we start pausing the game
+                        let pausetimeMaxTime = config.maxpausetime + Date.now(); // after the pause starts, the human player has  config.maxpausetime seconds before the game resumes
+                        // we send an ingame sendchat to notify the user of pause settings allowed, in human readable time
+                        conn_log(`The game has been paused, pausetime allowed : ${timespanToDisplayString(config.maxpausetime)}, remaining pauses allowed : ${timespanToDisplayString(config.maxpauses)}`);
+                        this.sendchat(`The game has been paused, pausetime allowed : ${timespanToDisplayString(config.maxpausetime)}, remaining pauses allowed : ${timespanToDisplayString(config.maxpauses)}`);
+
+                        let pausetimeCheck = function() {
+                            if (i >= pausetimeMaxTime) { // when all allowed pause time is consumed
```
roy7: The function won't know about pausetimeMaxTime from back when we set it because the function is called fresh at the time of the timeout. So we need to calculate anything needed inside this function.

However see my Discord comments which I'll pop into the conservation for the PR itself.

roy7:
Copy from the discord for easy reference/discussion.

On your pause thing what you want to do is a timeout based callback, which means you aren't doing loops or checking a variable. And variables set outside the callback won't be seen inside the callback function at the time it runs.

The nature of setting a timeout means after X seconds the code will run one time. So the high level idea is set a timeout equal to config.maxpausetime, which will run a small function that unpauses. Thus, if the game isn't already unpaused by then, it will auto unpause at the timeout.

To prevent stray timeouts running around, we will store the timeout in a variable saves in the Game object. When we can an unpause command from server, we clear any timeout (if one exists).

You can see how we save game connection timeouts in the this.connected_game_timeouts variable in connection.js.

Basically when see the pause come in, if we allow it, we'll setTimeout for an unpause function that will basically just this.resumeGame() and that's it. Maybe a chat message like "Max pause time exceeded, resuming game.".

Maybe before the first line ```if ((config.nopause || (config.nopauseranked && this.state.ranked) || (config.nopauseunranked && this.state.ranked == false) ``` we do something like

```if (clock.pause && !clock.pause.paused && this.clock_timeout) {
clearTimeout(this.clock_timeout);
this.clock_timeout = null;
}```

So if we get a clock tick and it says game isn't paused, we'll remove any pause timeout that is pending. Without testing or printing the clock messages to debug log I'm not sure offhand if this is exactly right or something else needs checked to properly see the unpause event come from server.

We'd also want to add into disconnect() in game.js something like

```if (this.clock_timeout) clearTimeout(this.clock_timeout);```

so we don't leave the timeout pending after the Game has been deleted from memory (it might still try to run later on and cause an error).

game.js
wonderingabout: indeed, i shouldnt put it in the function !!

wonderingabout:
still unfinished, but taking some of @roy7 's nice review into account :+1:

wonderingabout:
current error now : 
currenterror2

wonderingabout:
i'm also seeing this : 

https://github.com/wonderingabout/gtp2ogs/blob/hrr-display-rank--fix-rank-calculation/gtp2ogs.js#L1533

PR #121, opened 2019-02-07
Support max/min handicaps for autohandicap games (need server support?)

### LINKED ISSUE #28: Options for max and min rankedhandicap and 'auto' handicaps
Since automatic handicap provides handicap=-1 in the notification, we either need the API add info about our own handicap to the notification (so we can calculate the the expected real handicap at the time of notification) or fetch our user details each time a new challenge arrives.

Currently I just reject auto handicap in my production copy since it is a way away round the maxrankedhandicap setting.

wonderingabout:
this addressed and fixes #28 

added a dynamic behaviour when handicap is automatic and if automatic andicap is allowed (no --noautohandicap and co)

but i dont know how to express the bot's ranking 

how can i find this value ?

`player.ranking` and `notification.ranking` didnt work

@roy7 @Dorus @tokumotok

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -386,20 +386,61 @@ class Connection {
             return { reject: true, msg: "In your selected board size " + notification.width + "x" + notification.height + " (width x height), board HEIGHT (" + notification.height + ") is not allowed, please choose one of these allowed CUSTOM board HEIGHT values : " + config.boardsizeheight };
         }
 
-        if (config.noautohandicap && notification.handicap == -1) {
+        if (notification.handicap == -1 && config.noautohandicap) {
             conn_log("no autohandicap, rejecting challenge") ;
             return { reject: true, msg: "For easier bot management, automatic handicap is disabled on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
 	}
 
-        if (config.noautohandicapranked && notification.handicap == -1 && notification.ranked) {
+        if (notification.handicap == -1 && config.noautohandicapranked && notification.ranked) {
             conn_log("no autohandicap for ranked games, rejecting challenge") ;
             return { reject: true, msg: "For easier bot management, automatic handicap is disabled for ranked games on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
 	}
 
-        if (config.noautohandicapunranked && notification.handicap == -1 && !notification.ranked) {
+        if (notification.handicap == -1 && config.noautohandicapunranked && !notification.ranked) {
             conn_log("no autohandicap for unranked games, rejecting challenge") ;
             return { reject: true, msg: "For easier bot management, automatic handicap is disabled for unranked games on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
 	}
+
+        if (notification.handicap == -1 && !config.noautohandicap && !config.noautohandicapranked && !config.noautohandicapunranked) { 
+            // below is dynamic behaviour of autohandicap when handicap is automatic 
+            // and no --noautohandicap and co are set
+            let rankDifference = Math.abs(user.ranking - player.ranking);
+            // first, if ranked game, we eliminate > 9 rank difference
+            if (notification.ranked) {
+                if (rankDifference >= 9) {
+                    conn_log("Rank difference > 9 in a ranked game would be 10+ handicap stones, not allowed");
+                    return {reject: true, msg: "Rank difference too high to play a ranked handicap game, unranked is still possible"};
+                }
+            }
+            // then, after eliminating > 9 rank difference if ranked, we consider value of min-max handicap if set
+            // we eliminate all unwanted values, everything not forbidden is allowed
+            if (config.minhandicap || config.minhandicapranked || config.minhandicapunranked || config.maxhandicap || config.maxhandicapranked || config.maxhandicapunranked) {
+                if (config.minhandicap && (rankDifference < config.minhandicap)) {
+                    conn_log("Automatic handicap " + rankDifference + " stones is too low, minimum handicap stone number is " + config.minhandicap);
+                    return {reject: true, msg: "Automatic handicap " + rankDifference + " stones is too low, minimum handicap stone number is " + config.minhandicap + ", please increase the number of handicap stones in -custom handicap-"};
+                }
+                if (config.minhandicapranked && notification.ranked && (rankDifference < config.minhandicapranked)) {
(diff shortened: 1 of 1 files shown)


roy7:
I'm not sure that any challenge or connection result provides our rank. We might need to do a https://ogs.docs.apiary.io/#reference/user-specific-resources/user-details-resource/user-vitals call to check our rank at startup or something.

@anoek What would you think about adding the own plaeyer's rank into the challenge object as well? Then bots can use own rank and opponent rank to enforce min/max handicaps even if OGS sends out an auto handicap challenge.

wonderingabout:
@roy7 i thought about using a `--myfakerank` argument, for example `--myfakerank 7d`, but thats definitely not optimal

looking forward to being able to implement the ranking prorety in the challenger object !! @anoek

wonderingabout:
@Dorus 
that is the purpose of the current `notification.handicap`, for example if user chooses 5 handicap, minimum and maximum handicap can apply to that

but when handicap is automatic, instead of having the rank difference calculated to know handicap stone number, we only have a `-1`, undefined, so that currently doesnt work

so, my thoughts on this : even if we fix the autohandicap on the server side somehow, it would still be nice and useful to have the bot's ranking as a variable too

anoek:
FTR I'm going to add an estimated auto handicap in the challenge object, that should simplify this feature.

Also lets clean up the title of this pull request, just be clear and concise.

anoek:
Actually I just realized estimating auto handicap isn't really reliable since you can't do that if you have auto handicap and your color selection is random.

Is there a genuine need for this feature? I could see the need for a bot requiring only no handicap games, that's just detecting handicap == 0 so that shouldn't be a problem.

roy7:
Funny, I didn't realize auto handicap could be used with random color. Since the weaker player normally get black + the stones. If randomly the weaker player got white, what would happen? Normal game with zero handicap?

roy7:
What we're trying to solve is a bot that will, say, play with 2 handicap max but no more. We support that already for normal handicap stones. But if it is set to 'auto' we don't to accept or reject. We can just reject 'auto' but it's smoother for the human if we can accept auto if we believe the handicap will be within our allowed limited.

wonderingabout:
while we're at it i also play with my handicap bot

last time a 9d was forced to give my bot 2 stones !

@anoek @roy7 @Dorus https://online-go.com/game/16356789

this is because default color is set to auto

can we also add a way for bot admin to force his handicap bot to be color white (give stones is always white player on ogs)

ok i'll fix title

wonderingabout:
yes currently theres an issue with autohandicap

when handicap is automatic, it bypasses min and maxhandicap limits

for example if i set `--maxhandicap 0` then if handicap is automatic, the game will be played with several handicap stones and 0.5 komi

wonderingabout:
so this is addressing the bypass issue, by manually detecting number of handicap stones in "automatic handicap" to be within limits of min and maxhandicap set by botadmin

wonderingabout:
i was considering using `--myfakerank`, but definitely not optimal

Dorus:
@anoek Yes we need this, rejecting auto handicap is very frustrating for users as they will have to remake the challenge many times to find a setting the bot support. At the same time, many NN bots (esp leela) cannot support high handicap as they will surrender games at 92 moves. Lowering resign result in large trolling during the endgame, so that's not good either.

For handicap estimate. It would be enough to have the maximal number of handicap stones. Even if with random color, there could be 0 stones, if my bot could get 4+ stones if it dices white, it should still reject the challenge.

If we want to go fancy, we could also include who would receive the handicap stones `{ maxhandicap: 4, player: 'opponent' }` or something like that if the opponent would receive 4 stones (player id or something would work too)

wonderingabout:
@Dorus 

but we can just calculate the number of handicap stones based on our bot user ranking

the user ranking already exists for the opponent user, so i hope it can be added quite easily

also, maybe the bot ranking value will be useful some time in the future, while the number of handicap stones, i doubt

Dorus:
Yes, (estimated) opponent rank would work too.

Dorus:
Btw on the gtp2ogs part, it would be nice if we could set a max handicap against and a max handicap with.

I wouldn't mind if my bot plays with 9 stones, but since Leela is totally broken if she play against 4+ stones (and even with 2 it's way more likely to fall for ladders or surrender at move 92), but playing with 9 stones is not really a problem, other than potentially slack moves.

wonderingabout:
@Dorus 

yes, definitely

more generally, i think we bot admins want to have the options to always play white if we want (i.e. we always give stones to our opponent even if he is higher ranked than us)

see : #127 for the issue i had against a 9d

so no matter how we look at it, we really need our bot's ranking available in the object too @anoek 

big thanks

Dorus:
yes you already said that 4 days ago, i'm not sure why you have the repeat.

But what i said is the opposite. I want my bot to always accept handicap games as black, even with +9 handicap stones, but only with max n handicap as white, say max -2 or -4.

It would be nice if we could set both a min (negative) handicap to play against, and a max (positive) handicap to play with. I would set it to [-2, +9] while you would set it to [-4, 0].

off-topic
---
Also i do not think you can force your bot to play white with random color. At best you can auto surrender if you dislike the color. But tbh i would find a bot that does that a poor sport. If somebody wants to play a zero handicap game vs a bot, they need to set handicap to none, not to auto. Also another issue is that bot ranking is very inaccurate. Both because there are very few 9d players on ogs, people cheat vs the bots, and bots are often vulnerable to certain play, while they overall pay at 12d or better. But this is nothing something we can solve.

wonderingabout:
this is easily configurable if we know our bot's ranking : 

` if (rankDifference >=0 && ((notification.handicap == -1 ) || (notification.handicap > 0))) `

wonderingabout:
@Dorus we dont need nor want to force,

we only reject game if opponent is not black color

we can guess the player's color indirectly from the settings in my example above

however if we have a `notification.user.color` (maybe it already exists ?), then its even easier (because theres the case where our opponent set color black, then we'll still reject game because we cant detect his color)

wonderingabout:
@Dorus 

more generally for your request we can code it easily, we just need 2 new features as you are aware : 
- 1) `player.ranking` (our bot's ranking)
- 2) `notification.user.color` (know player color in challenge settings)

with these 2 data, we can program it easily as you know better than me :+1:

wonderingabout:
as for the issue of superhuman bots, we already know that humans cant win against them
also, they are very bad at handicap, so it cant be used to measure their ranking

so we only have one thing to do, obviously : 
disable ranked games for all superhuman bots (and allow it only between such superhuman bots : 

example allow ranked game leela zero vs phoenixgo, but do not allow ranked leela zero vs random 9d human player, this random 9d can always play unranked against superhuman bots, or ranked against human 9d or less)

Issue #59, opened 2018-03-18
Simultaneous games question


lemonsqueeze:
Hi, i'm trying hook up Pachi to gtp2ogs, see how it does on ogs.
So far it's working fine for single games, but looks like i'm having an issue with simultaneous games.

How is it supposed to work at the moment ? gtp2ogs spawns one engine instance per simultaneous game ?
(trying live games atm, running with --persist)

Looks like this is what's happening, my engine script gets spawned again when the 2nd game starts and it's not happy about it. Maybe i need PR #44 to use a single instance instead ? But this only works for correspondence games i guess...

Otherwise i can fix my side to handle multiple instances, but i need a way to set max number of simultaneous games (if it spawns 30 instances it's going to be an issue ...)

roy7:
Hi there! Yes OGS's model is basically that a bot plays all the games at once that people want. (I think the KGS model might be bots play one game at a time only, I'm not sure.)

If you want to stick to only slower time settings you could see about changing --corrqueue in some way to try and only work on one game at a time for cpu/gpu purposes (like turn it on but use it for all games not just correspondence games). If you turn off --persist, it'll restart your bot fresh for each move. I added --persist since I wanted the prior tree search in Leela to stick around from move to move.

There's no easy way atm to "reject matches if a match is under way" although it has been requested before. I'm not sure if Tokumoto ever added that in his own bots. It has been requested before but I never wrote it up myself.

lemonsqueeze:
Thanks, this makes sense. Yes, i'm a bit new to the ogs model.

Managed to get simultaneous games working !
For now i let it spawn multiple instances and they have a queue to decide whose turn it is to run.
This works pretty well with a few simul games, but it becomes an issue when too many people start hammering the thing at the same time. I guess it's the same with Leela, how do you deal with it ? Just require big enough time settings to avoid timeouts ?

I'd rather avoid running without --persist, starting a new instance for each move is quite slow. But then if it solves all the timeout / concurrency issues maybe it's worth looking into...

I'm thinking maybe best is to signal gtp2ogs to start/stop accepting new games. I could check system load from time to time and restart it with --rejectnew when it gets too much (--rejectnew isn't going to affect active games, is it ?)

roy7:
--rejectnew won't affect ongoing games, it just rejects new challenges. But you can go to the accept challenge part of the code and maybe add a check for your system load so you reject if load is too high.

[...]

PR #123, opened 2019-02-08
Rework console warnings + add defaults + no 3 of same family behaviour + minrankranked and co + maxconnectedgames maxconnectedgamesperuser + fix missing exports

wonderingabout:
Many new changes in this PR, some are important :

general look here : https://github.com/wonderingabout/gtp2ogs/tree/rework-console-minrankranked-and-co 

- removed console warnings for maxmaintime and co "setting not detected"
- instead, using default values hassle-free, large and friendly for rescue value purposes
- change behaviour if 3 arguments of same family are used : for arguments that have a general+ranked+unranked possible values, new behaviour is now, for example for `--maxmaintime` :  general value (default or manually inputted by bot admin) will be taken into account only if no specific value is used (ranked and/or unranked); else bot admin will have to set up both ranked and unranked values desired and general value will not be taken into account (added notes about that on readme)
- added `--minrankranked` and co : if bot admin wants to specify different minrank and maxrank for ranked and unranked games
- renamed `--maxactivegames` and `--maxtotalgames` to the clearer and more powerful `--maxconnectedgames` and `--maxconnectedgamesperuser` : these names are clearer but also more powerful : in the future, someone may want to add a maxtotalgames per bot and/or per user which is something different (max number of total games in the list on ogs profile, see gnugo page for example : it has a lot of games in the list but gnugo is not connected to almost any of them, bot admins may want to have the option to regulate that in the future) @lemonsqueeze 
- updated readme and ISSA to take all that into account

@roy7 @Dorus @windo @lemonsqueeze

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,12 +143,12 @@ Then, the following options are placed in the above ```<gtp2ogsarguments>```
 section. Put a space in between options when there are more than one.  
 Also put a space in between the option and the parameter, for example :
 
-  ```--startupbuffer 2 --boardsize 13,19 --noclock --unrankedonly --maxactivegames 1 --maxmaintime 1200 --ban UserX,playerY ---maxperiodsranked 5```
+  ```--startupbuffer 2 --boardsize 13,19 --noclock --unrankedonly --maxconnectedgamesperuser 1 --maxconnectedgames 10 --maxmaintime 1200 --ban UserX,playerY ---maxperiodsranked 5```
   
 or with ISSA (intuitive semi-syllabic aliases) (see [notes G-](/NOTES.md#g-) for details), 
 the same example becomes :
   
-  ```--sb 2 --bb 13,19 --nc --uo --1ag 1 --1mt 1200 --b UserX,playerY --1pr 5```
+  ```--sb 2 --bb 13,19 --nc --uo --1cgpu 1 --1cg 10 --1mt 1200 --b UserX,playerY --1pr 5```
   
 note : some gtp2ogsarguments have default so they are enabled even if you don't 
 specify them, such as `--komi` which default is automatic even if you dont specify it !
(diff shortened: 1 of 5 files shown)


wonderingabout:
update : currently not working for some reason, most likely when i renamed maxactivegames to maxtotalgames

will debug next time

wonderingabout:
tested to work on beta ogs ingame

roy7:
> correspondence games are currently included in the connected games count if you use `--persist`

I don't think this is how it works? I believe every game in progress will be in the connected games list, whether it is your turn or not. --persist won't change that. --persist would change the # of engines in memory/active though, if you go through the connected games list to count how many have valid game states? I'm just going from memory here.

README.md
```diff
@@ -241,10 +250,10 @@ are banned from playing unranked game
 byoyomi,simple,canadian,absolute,none)
 
   ```--minmaintime```  or ```--0mt``` Minimum seconds of main time (rejects time 
-control simple and none)
+control simple and none) (default 60)
 
   ```--maxmaintime```  or ```--1mt``` Maximum seconds of main time (rejects time 
-control simple and none)
+control simple and none) (default 1800)
```
roy7: This is only 30 minutes. How about making default max main time 2 or 4 hours. The default on max time is more of a "prevent unintended unlimited time". If people want less than a high amount they can lower in settings. 30 minutes feels too low for default max, since often people do in fact play bots with longer clocks?

wonderingabout: i see, this is why i wanted to make separate time settings for everygame speed : 
- `--maxmaintimeblitzlive 60*60`
- `--maxmaintimecorr 60*60*24*7`

i think every speed should have its own time speed @roy7 
more than 1 hour main time is not a live game anymore (blitz and live are kidna same so same category)
if main time is more than 1 hour, i suggest the human player should play a corr game instead

as you said, bot admin is free to increase to `--maxmaintimeblitzlive 60*60*4` if want @roy7 
i think this makes more sense 

see #125 
also, with a config file we'd be able to do math like that automatically which is also a nice advantage

to sum up, i can increase to 1 hour, but more than that should be a corr game by default, unless bot admin wants to manually change that
(after main time there are still periods : default is 1 hour + 20 x 2 min if i remember correctly, theres plenty of time for a live game, much more than on foxwq or tygem for example)

wonderingabout:
> > correspondence games are currently included in the connected games count if you use `--persist`
> 
> I don't think this is how it works? I believe every game in progress will be in the connected games list, whether it is your turn or not. --persist won't change that. --persist would change the # of engines in memory/active though, if you go through the connected games list to count how many have valid game states? I'm just going from memory here.

maybe i confused `--persist` with corr settings, because i remember there was an issue here in #59

README.md
Dorus: I agree with Roy here. Default should be something large like 2-4 hours. If bot admins find that too much, they can decrease it.

wonderingabout:
so to sum up i'll increase maxmaintimelive to 3600

and more importantly, i'll enhance the UHMAEAT to support separate blitz,live, and corr timesettings (we'll have 3 UHMEAT for maintime, and 3 for periodtime)

for example : 

maxmaintimeblitz 60*5
maxmaintimelive 60*60
maxmaintimecorr 60*60*24*7

maxperiodtimebltz 30
maxperiodtimelive 60*2
maxperiodtimecorr 60*60*24*3

@Dorus @roy7

README.md
wonderingabout: ok, i'll increase it to 2 hours then, but remember that there is still periodtime on top of that

and with maxmaintime 2 hours, it means corr games are not possible, all the more reason to add a specific setting for blitz, live corr, see my messages in https://github.com/online-go/gtp2ogs/pull/123#issuecomment-462112142

wonderingabout:
reverted to 7200 default in #124 

let's continue the discussion there :+1:

PR #132, opened 2019-02-17
fix whitespace and add vim markers for formatting

windo:
I had messed up some 4 vs 2 spaces earlier so adding markers for vim to force corredt indentation settings. Also fixing incorrect indentation in a few places.


roy7:
Ok only conflict was a change to global ban vs ranked/unranked ban from earlier PR so I fixed that conflict and did the merge.

Issue #139, opened 2019-02-21
server disconnects can cause games to get stuck


windo:
It looks like if a bot is in the middle of the move and the OGS server is restarted, the bot will get stuck and never move. The gtp2ogs logs in this case look like:

```
Feb 21 10:06:56   game.js:474                    [Game 16674926]  Move 23   genmove white
Feb 21 10:06:56   bot.js:130                     [26894] Generating move for game 16674926
Feb 21 10:06:56   bot.js:130                     [26894] >>> genmove white
Feb 21 10:06:56   bot.js:130                     [26894] <<< =
Feb 21 10:06:56   connection.js:1659             #  Disconnected from server
Feb 21 10:06:56   connection.js:1659             #  disconnectFromGame 16674226
Feb 21 10:06:56   bot.js:130                     [24595] Stopping bot
Feb 21 10:06:56   bot.js:130                     [24595] Attempting to send to dead bot: quit
Feb 21 10:06:56   game.js:474                    [Game 16674226]  Move 151  Disconnecting from game.
Feb 21 10:06:56   connection.js:1659             #  disconnectFromGame 16674875
Feb 21 10:06:56   bot.js:130                     [24246] Stopping bot
Feb 21 10:06:56   bot.js:130                     [24246] Attempting to send to dead bot: quit
Feb 21 10:06:56   game.js:474                    [Game 16674875]  Move 134  Disconnecting from game.
Feb 21 10:06:56   connection.js:1659             #  disconnectFromGame 16674899
Feb 21 10:06:56   bot.js:130                     [26314] Stopping bot
Feb 21 10:06:56   bot.js:130                     [26314] Attempting to send to dead bot: quit
Feb 21 10:06:56   game.js:474                    [Game 16674899]  Move 71   Disconnecting from game.
Feb 21 10:06:56   connection.js:1659             #  disconnectFromGame 16674926
Feb 21 10:06:56   bot.js:130                     [26894] Stopping bot
Feb 21 10:06:56   bot.js:130                     [26894] Attempting to send to dead bot: quit
Feb 21 10:06:56   game.js:474                    [Game 16674926]  Move 23   Disconnecting from game.
Feb 21 10:06:56   bot.js:130                     [24246] Bot exited
Feb 21 10:06:56   bot.js:130                     [24595] Bot exited
Feb 21 10:06:56   bot.js:130                     [26314] Bot exited
Feb 21 10:06:56   bot.js:130                     [26894] Bot exited
[...]

PR #142, opened 2019-02-24
Possibly fix hanging games on server restart

windo:
This intends to cover #139. The hanging games happen every few days so I can't vouch this fixes the problem for sure.

In particular, I'm not sure if `game/disconnect` followed by `game/connect` is the best way to get a `gamedata` to be sent again.

Also not sure if 1s is the right timeout. The timeout should probably be short (because a clock might be ticking for the game) so it felt like a good initial guess to me.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -228,24 +228,26 @@ class Connection {
             let game = this.connected_games[game_id];
             let msg = [];
             msg.push('game_id=' + game_id + ':');
-            if (game.state == null) {
+            if (game.state === null) {
                 msg.push('no_state');
                 conn_log(...msg);
                 continue;
             }
             msg.push('black=' + game.state.players.black.username);
             msg.push('white=' + game.state.players.white.username);
-            if (game.state.clock.current_player == this.bot_id) {
+            if (game.state.clock.current_player === this.bot_id) {
                 msg.push('bot_turn');
             }
             let idle_time = (Date.now() - game.state.clock.last_move) / 1000;
             msg.push('idle_time=' + idle_time + 's');
-            if (game.bot == null) {
+            if (game.bot === null) {
                 msg.push('no_bot');
                 conn_log(...msg);
                 continue;
             }
-            msg.push('bot.proc.pid=' + game.bot.proc.pid);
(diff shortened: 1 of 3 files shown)


connection.js
```diff
@@ -218,6 +222,34 @@ class Connection {
             }
         }
     };
+    dumpStatus() {
+        conn_log('Dumping status of all connected games');
+        for (let game_id in this.connected_games) {
+            let game = this.connected_games[game_id];
+            let msg = [];
+            msg.push('game_id=' + game_id + ':');
+            if (game.state == null) {
```
roy7: Since we're doing a general style overhaul to use === and !== everywhere would you change this over to ===? Unless there's a reason we shouldn't do the overhaul?

connection.js
```diff
@@ -218,6 +222,34 @@ class Connection {
...
+                msg.push('no_state');
+                conn_log(...msg);
+                continue;
+            }
+            msg.push('black=' + game.state.players.black.username);
+            msg.push('white=' + game.state.players.white.username);
+            if (game.state.clock.current_player == this.bot_id) {
```
roy7: ===

connection.js
```diff
@@ -218,6 +222,34 @@ class Connection {
...
+                msg.push('bot_turn');
+            }
+            let idle_time = (Date.now() - game.state.clock.last_move) / 1000;
+            msg.push('idle_time=' + idle_time + 's');
+            if (game.bot == null) {
```
roy7: ===

windo: Done

connection.js
windo: Done

connection.js
windo: Done

windo:
IDK how the etiquette works on github - am I supposed to mark changes as done or leave it for the reviewer to mark the changes done once they are satisfied?

roy7:
It's really just the 4-5 of us so it's very informal around here. I believe you are welcome to resolve them and also I think once those code lines are changed it automatically hides them as 'outdated'.

PR #144, opened 2019-02-24
code optimization : using let e of arrayValues

wonderingabout:
@roy7 @Dorus @windo 

`let i of` 
is not clear, because `i` generally refers to the index, not the element

also replacing :
`let value of values` 
with : 
`let e of values`
for easier code maintainance and clearer meaning

didnt test it, but it should work unless i did a mistake :+1: 

eventually we would want to replace these with functions, but this is a good start i think :+1: 

we can take care of all the `==` and `!==` in the code in a separate PR @Dorus  :+1:

### DIFF
--- a/config.js
+++ b/config.js
@@ -387,20 +387,20 @@ exports.updateFromArgv = function() {
     }
 
     if (argv.ban) {
-        for (let i of argv.ban.split(',')) {
-            exports.banned_users[i] = true;
+        for (let e of argv.ban.split(',')) {
+            exports.banned_users[e] = true;
         }
     }
 
     if (argv.banranked) {
-        for (let i of argv.banranked.split(',')) {
-            exports.banned_ranked_users[i] = true;
+        for (let e of argv.banranked.split(',')) {
+            exports.banned_ranked_users[e] = true;
         }
     }
 
     if (argv.banunranked) {
-        for (let i of argv.banunranked.split(',')) {
-            exports.banned_unranked_users[i] = true;
+        for (let e of argv.banunranked.split(',')) {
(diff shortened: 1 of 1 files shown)


config.js
```diff
@@ -387,20 +387,20 @@ exports.updateFromArgv = function() {
...
-            exports.banned_ranked_users[i] = true;
+        for (let e of argv.banranked.split(',')) {
+            exports.banned_ranked_users[e] = true;
         }
     }
 
     if (argv.banunranked) {
-        for (let i of argv.banunranked.split(',')) {
-            exports.banned_unranked_users[i] = true;
+        for (let e of argv.banunranked.split(',')) {
```
windo: These seem good.

config.js
```diff
@@ -534,128 +534,128 @@ exports.updateFromArgv = function() {
     }
 
     if (argv.boardsize) {
-        for (let boardsize of argv.boardsize.split(',')) {
-            if (boardsize == "all") {
+        for (let e of argv.boardsize.split(',')) {
```
windo: Not sure if it is an improvement if we already named the variable. I'd say `boardsize` is clearer than `e`.

roy7:
I agree if a variable is already clearly named just leave it be.

wonderingabout:
ok, but then shouldnt we name all variables, and never use `e` ?

while it may be better to use name value, it adds more code maintainance in the current code
(writing general/ranked/unranked many times for boardsize, speed, komi, timecontrols)

wonderingabout:
now that this is all clear, i'll submit first the == and != PR separately first, will add this one top of it

PR #146, opened 2019-02-24
Handle crash sendchat

wonderingabout:
moving this out of #145 

so this PR is on top of #145 

see #145 for talks details

@roy7 @Dorus @windo 
tested to work, but currently the log and senchats are not what we want (wrong order of messages + two resigning messages when we resign only once)

### DIFF
--- a/game.js
+++ b/game.js
@@ -250,7 +250,8 @@ class Game {
 
         if (this.bot_failures >= 5) {
             // This bot keeps on failing, give up on the game.
-            this.log("Bot has crashed too many times, resigining game");
+            this.log("Bot has crashed too many times, resigning game");
+            this.sendChat("Bot has crashed too many times, resigning game"); // we notify user of this in ingame chat
             this.socket.emit('game/resign', this.auth({
                 'game_id': this.state.game_id
             }));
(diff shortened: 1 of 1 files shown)


roy7:
#145 is merged, can you do a git pull?

wonderingabout:
yes ok @roy7 

but here is a note for @windo to remember to look why log acts in a weird way :+1:

wonderingabout:
done @roy7 :+1:

game.js
```diff
@@ -299,7 +300,8 @@ class Game {
         }
 
         if (!this.ensureBotStarted(botError)) {
-            this.log("Failed to start the bot, can not make a move");
+            this.log("Failed to start the bot, can not make a move, trying to restart");
+            this.sendChat("Failed to start the bot, can not make a move, trying to restart"); // we notify user of this in ingame chat
```
windo: I don't think the user needs to know about this. But don't particularly mind.

wonderingabout: @windo 
imagine if our bot is slow to start/restart

then the user may have to wait many minutes without knowing what is going on

i think in that case, waiting say 30 seconds instead with a message after 30 seconds is much better :+1:

PR #147, opened 2019-02-25
eslint: require eqeqeq

windo:
I just created a plain eslint config that currently just enforces
eqeqeq. We can add other coding conventions as specific rules into the
config later.

In particular, we should probably enable the `eslint:recommended` base
configuration that enables quite a few common options:

https://eslint.org/docs/rules/

But can't do it right off the bat because that requires a few more
cleanups before it would run cleanly on the codebase.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,0 +1,11 @@
+env:
+  node: true
+  es6: true
+globals:
+  Atomics: readonly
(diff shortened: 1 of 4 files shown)


roy7:
Ok I'll merge this and we can do the other eslint items as you direct later on. :)

PR #150, opened 2019-02-26
fix --komi auto depreciation warning message

wonderingabout:
@roy7 @Dorus @windo 

the previous one was displaying the warning
everytime, because of the unappropriate use
of .includes

replaced with a simpler and correct code now

tested to work :+1:

PR #151, opened 2019-02-26
Major Enhancement of console warnings display

wonderingabout:
general look here : https://github.com/wonderingabout/gtp2ogs/blob/enhance-console-messages-display/config.js

this PR is on top of #150 

now using functions to avoid needless and unreliable
repetitions

also added an [OK!] or [ERROR(S)] display result
(3 new indicators)

also pushing many fixes in in this commit as well

currently, error : parameter is not defined

newfunctionsnotdefined

i'm not experienced with functions,
what should i do to fix this ?

@roy7 @Dorus @windo

### DIFF
--- a/config.js
+++ b/config.js
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
         process.exit();
     }
 
-    // console : greeting //
-
+    // console : boot message //
     console.log("\nYou are using gtp2ogs version 6.0\n- For changelog or latest devel updates, please visit https://github.com/online-go/gtp2ogs/tree/devel\n");
 
-    // console : warnings //
+    /*** console : DEPRECIATIONS ***/
+    console.log("-----DEPRECIATIONS-----");
+    this.indicatorConsoleDepreciation = false;
 
-    // - warning : dont use 3 settings of the same family (general, ranked, unranked) at the same time
-    if (argv.maxhandicap && (argv.maxhandicapranked || argv.maxhandicapunranked)) {
-        console.log("Warning: You are using --maxhandicap in combination with --maxhandicapranked and/or --maxhandicapunranked.\nUse either --maxhandicap alone, OR --maxhandicapranked with --maxhandicapunranked.\nBut don't use the 3 maxhandicap arguments at the same time.");
-    }
-
-    if (argv.minhandicap && (argv.minhandicapranked || argv.minhandicapunranked)) {
-        console.log("Warning: You are using --minhandicap in combination with --minhandicapranked and/or --minhandicapunranked.\nUse either --minhandicap alone, OR --minhandicapranked with --minhandicapunranked. \nBut don't use the 3 minhandicap arguments at the same time.");
-    }
-
-    if (argv.noautohandicap && (argv.noautohandicapranked || argv.noautohandicapunranked)) {
-        console.log("Warning: You are using --noautohandicap in combination with --noautohandicapranked and/or --noautohandicapunranked.\nUse either --noautohandicap alone, OR --noautohandicapranked with --noautohandicapunranked.\nBut don't use the 3 noautohandicap arguments at the same time.");
-    }
-
-    if (argv.maxmaintime || argv.maxmaintimeranked || argv.maxmaintimeunranked || argv.minmaintime || argv.minmaintimeranked || argv.minmaintimeunranked) {
-        console.log("Warning: --min/max*maintime*+/-ranked/unranked is not supported anymore\n Use --min/max*maintime*blitz/live/corr*+/-ranked/unranked");
-    }
-
-    if (argv.maxperiodtime || argv.maxperiodtimeranked || argv.maxperiodtimeunranked || argv.minperiodtime || argv.minperiodtimeranked || argv.minperiodtimeunranked) {
-        console.log("Warning: --min/max*periodtime*+/-ranked/unranked is not supported anymore\n Use --min/max*periodtime*blitz/live/corr*+/-ranked/unranked");
-    }
-
-    if (argv.maxperiods && (argv.maxperiodsranked || argv.maxperiodsunranked)) {
-        console.log("Warning: You are using --maxperiods in combination with --maxperiodsranked and/or --maxperiodsunranked.\nUse either --maxperiods alone, OR --maxperiodsranked with --maxperiodsunranked.\nBut don't use the 3 maxperiods arguments at the same time.");
-    }
-
-    if (argv.minperiods && (argv.minperiodsranked || argv.minperiodsunranked)) {
-        console.log("Warning: You are using --minperiods in combination with --minperiodsranked and/or --minperiodsunranked.\nUse either --minperiods alone, OR --minperiodsranked with --minperiodsunranked.\nBut don't use the 3 minperiods arguments at the same time.");
-    }
-
-    if (argv.minrank && (argv.minrankranked || argv.minrankunranked)) {
-        console.log("Warning: You are using --minrank in combination with --minrankranked and/or --minrankunranked. \n Use either --minrank alone, OR --minrankranked with --minrankunranked.\nBut don't use the 3 minrank arguments at the same time.");
-    }
-
-    if (argv.maxrank && (argv.maxrankranked || argv.maxrankunranked)) {
-        console.log("Warning: You are using --maxrank in combination with --maxrankranked and/or --maxrankunranked. \n Use either --maxrank alone, OR --maxrankranked with --maxrankunranked.\nBut don't use the 3 maxrank arguments at the same time.");
-    }
-
-    if (argv.nopause && (argv.nopauseranked || argv.nopauseunranked)) {
-        console.log("Warning: You are using --nopause in combination with --nopauseranked and/or --nopauseunranked. \n Use either --nopause alone, OR --nopauseranked with --nopauseunranked.\nBut don't use the 3 nopause arguments at the same time.");
-    }
-
-    if (argv.komi) {
-        if (argv.komi.includes(`auto`)) {
-            console.log("Warning: /--komi auto/ has been renamed to /--komi automatic/\n");
-            // we skip a line here, not below, because this argv may be undefined if not used by bot admin
+    function depreciatedArgumentIsUsedNewArgumentToUseInstead(depreciatedArgumentUsed, newArgumentToUseInstead, indicatorDepreciation) {
+        if (depreciatedArgumentUsed) {
+            indicatorDepreciation = true;
+            console.log(`Warning: --${depreciatedArgumentUsed} is no longer supported. Use --${newArgumentToUseInstead} instead.\nRead latest up to date available gtp2ogs arguments in OPTIONS-LIST (devel branch) for details : https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md`);
         }
     }
 
-    if (argv.komiranked) {
-        if (argv.komiranked.includes(`auto`)) {
-            console.log("Warning: /--komiranked auto/ has been renamed to /--komiranked automatic/\n");
-            // we skip a line here, not below, because this argv may be undefined if not used by bot admin
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.botid, argv.username, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.bot, argv.username, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.id, argv.username, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.minrankedhandicap, argv.minhandicapranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.minunrankedhandicap, argv.minhandicapunranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.maxrankedhandicap, argv.maxhandicapranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.maxunrankedhandicap, argv.maxhandicapunranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.maxtotalgames, argv.maxconnectedgames, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.maxactivegames, argv.maxconnectedgamesperuser, this.indicatorConsoleDepreciation);
+
+    function depreciatedArgumentIsUsedNewArgumentToUseInsteadUpToThree(depreciatedArgumentUsed, newArgumentToUseInsteadOne, newArgumentToUseInsteadTwo, newArgumentToUseInsteadThree, indicatorDepreciation) {
+        if (depreciatedArgumentUsedOne) {
+            indicatorDepreciation = true;
+            console.log(`Warning: --${depreciatedArgumentUsed} is no longer supported. Use --${newArgumentToUseInsteadOne} and/or --${newArgumentToUseInsteadTwo} and/or --${newArgumentToUseInsteadThree} instead.\nDefaults are provided for all the new arguments, so you dont need to use them all\nRead latest up to date available gtp2ogs arguments in OPTIONS-LIST (devel branch) for details : https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md`);
         }
     }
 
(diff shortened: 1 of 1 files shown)


config.js
```diff
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
...
-    }
-
-    if (argv.maxperiods && (argv.maxperiodsranked || argv.maxperiodsunranked)) {
-        console.log("Warning: You are using --maxperiods in combination with --maxperiodsranked and/or --maxperiodsunranked.\nUse either --maxperiods alone, OR --maxperiodsranked with --maxperiodsunranked.\nBut don't use the 3 maxperiods arguments at the same time.");
-    }
-
-    if (argv.minperiods && (argv.minperiodsranked || argv.minperiodsunranked)) {
-        console.log("Warning: You are using --minperiods in combination with --minperiodsranked and/or --minperiodsunranked.\nUse either --minperiods alone, OR --minperiodsranked with --minperiodsunranked.\nBut don't use the 3 minperiods arguments at the same time.");
-    }
+    /*** console : DEPRECIATIONS ***/
```
windo: Should be `DEPRECATIONS` (note the lack of 'i').

Depreciate is a very similar word with a similarish meaning, but in this case the normal term is [deprecate](https://en.wikipedia.org/wiki/Deprecation).

config.js
```diff
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
...
+    console.log("-----DEPRECIATIONS-----");
+    this.indicatorConsoleDepreciation = false;
```
windo: `let deprecationIndicator = false;`

There should be no `this` reference in scope at this point. We are setting variables in `exports` which is a node-specific module-level variable, but there is no particular object we are working in.

config.js
```diff
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
...
-
-    if (argv.nopause && (argv.nopauseranked || argv.nopauseunranked)) {
-        console.log("Warning: You are using --nopause in combination with --nopauseranked and/or --nopauseunranked. \n Use either --nopause alone, OR --nopauseranked with --nopauseunranked.\nBut don't use the 3 nopause arguments at the same time.");
-    }
-
-    if (argv.komi) {
-        if (argv.komi.includes(`auto`)) {
-            console.log("Warning: /--komi auto/ has been renamed to /--komi automatic/\n");
-            // we skip a line here, not below, because this argv may be undefined if not used by bot admin
+    function depreciatedArgumentIsUsedNewArgumentToUseInstead(depreciatedArgumentUsed, newArgumentToUseInstead, indicatorDepreciation) {
```
windo: I'm pretty sure passing in `indicatorDepreciation` here does not work as you intend it to work:

```
let a = false;
function b(a) {
	a = true;
}
b(a);
display.log(a);
```
Results in: `false`

On the other hand, having the function refer to the same variable as captured from the scope of the function definition:

```
let a = false;
function b() {
	a = true;
}
b();
display.log(a);
````

Results in: `true`.

Read about closures if this sounds unexpected, for example [this](https://www.w3schools.com/js/js_function_closures.asp).

Comment (windo) [inline config.js]: Also, since this function is only used locally, we don't have to give it quite as long a name - in my opinion.

Generally with variable naming - the things you use only on a few nearby lines can have a single-character name, things you use in a function an abbreviation. Functions, object attributes and other things used across a whole module/file should probably have clear names and global cross-module names (or APIs even) should be extra careful to be clear.

But do always ask if a part of the name can be removed without losing clarity. I think in this case for example `depreciatedArgumentWarning()` is sufficient and since it is only used in this function, you might be able to make it shorter still without causing any confusion.

File: config.js [hunk 4]
```diff
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
...
+        if (depreciatedArgumentUsed) {
```
windo: In this case I think for example the `Used` part of the variable name can be dropped. And since it is a local variable, you could probably shorten in it further (though this is more a matter of taste), maybe `deprecatedArg`.

config.js
```diff
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
...
+            indicatorDepreciation = true;
+            console.log(`Warning: --${depreciatedArgumentUsed} is no longer supported. Use --${newArgumentToUseInstead} instead.\nRead latest up to date available gtp2ogs arguments in OPTIONS-LIST (devel branch) for details : https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md`);
```
windo: Unless I'm misunderstanding, I think `depreciatedArgumentUsed` is the **value** of the argument (looking below, for example `argv.botid`), not the name of the argument (which would be the string `"botid"`). And in the warning message, you want the name.

config.js
```diff
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
...
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.id, argv.username, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.minrankedhandicap, argv.minhandicapranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.minunrankedhandicap, argv.minhandicapunranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.maxrankedhandicap, argv.maxhandicapranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.maxunrankedhandicap, argv.maxhandicapunranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.maxtotalgames, argv.maxconnectedgames, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInstead(argv.maxactivegames, argv.maxconnectedgamesperuser, this.indicatorConsoleDepreciation);
+
+    function depreciatedArgumentIsUsedNewArgumentToUseInsteadUpToThree(depreciatedArgumentUsed, newArgumentToUseInsteadOne, newArgumentToUseInsteadTwo, newArgumentToUseInsteadThree, indicatorDepreciation) {
```
windo: I think in this case you want something like:

```
function deprecatedArgumentAlternatives(deprecatedArg, alternatives) {
  console.log("Warning: --${deprecatedArg} is no longer supported. Use one of",
      alternatives.map(x => ("--" + x)).join(" ,"),
      "instead.";
  )
}
```

To avoid having separate helper functions for arguments with different numbers of alternatives.

config.js
```diff
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
...
+    depreciatedArgumentIsUsedNewArgumentToUseInsteadUpToThree(argv.minperiodtimeranked, argv.minperiodtimeblitzranked, argv.minperiodtimeliveranked, argv.minperiodtimecorrranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInsteadUpToThree(argv.minperiodtimeunranked, argv.minperiodtimeblitzunranked, argv.minperiodtimeliveunranked, argv.minperiodtimecorrunranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInsteadUpToThree(argv.maxperiodtime, argv.maxperiodtimeblitz, argv.maxperiodtimelive, argv.maxperiodtimecorr, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInsteadUpToThree(argv.maxperiodtimeranked, argv.maxperiodtimeblitzranked, argv.maxperiodtimeliveranked, argv.maxperiodtimecorrranked, this.indicatorConsoleDepreciation);
+    depreciatedArgumentIsUsedNewArgumentToUseInsteadUpToThree(argv.maxperiodtimeunranked, argv.maxperiodtimeblitzunranked, argv.maxperiodtimeliveunranked, argv.maxperiodtimecorrunranked, this.indicatorConsoleDepreciation);
+
+    function isArgumentArrayEqualToAuto(argumentToTest, indicatorDepreciation) {
```
windo: I would just call this `checkForDeprecatedKomi()` - this is a special purpose function and it's best to name it for what is the task that this function is supposed to accomplish. Especially as it also checks for `null`.

config.js
```diff
@@ -238,114 +238,130 @@ exports.updateFromArgv = function() {
...
-        console.log("Warning: --id alias is no longer supported. Use --username instead.");
-    }
-
-    if (argv.minrankedhandicap || argv.minunrankedhandicap || argv.maxrankedhandicap || argv.maxunrankedhandicap) {
-        console.log("Warning: --min/max*+/-ranked/unranked*handicap argument is no longer supported.\nUse --min/max*handicap*+/-ranked/unranked instead.");
-    }
-
-    if (argv.maxtotalgames) {
-        console.log("Warning: --maxtotalgames argument has been renamed to --maxconnectedgames. Use --maxconnectedgames instead.");
+    function generalArgumentIsUsedWithSpecificAtTheSameTime(generalArgument,rankedArgument,unrankedArgument, indicatorWarningsGeneralRankedUnranked) {
```
windo: I like this, but maybe it could be taken even further and work similarly as my `alternatives` suggestion above? So the helper function signature would be:

```
function checkConflictingArgs(generalArg, alternatives) {
    ...
}
```

So you could give it any number of arguments that conflict with the general argument.

windo:
Oh, and as for the failure on line 266 (`depreciatedArgumentUsedOne`), I think there is a difference in the name of the function argument and the name you are using to refer to it on that line?

wonderingabout:
> So I thought maybe some extra discussion would be helpful.
> 
> But I do think this PR looks good even without touching it up too much.

of course :1st_place_medal:  :1st_place_medal:  :1st_place_medal: 
i am very thankful and really looking forward to more advanced things, please dont hold back on that and give me as many things to start looking for as you think are helpful :+1:  big thanks @windo 

oh and dont hesitate to criticize as much as you want, i welcome that, no need to hold back lol :+1: (i only care about the how, not the what) :+1:

config.js
wonderingabout: oh i see :+1: 

i thought we were in the config object, so i did same as what you did for universalMaintime

i dont know about nodejs, but i see its clear :+1:

config.js
wonderingabout: oh
in french we use only "deprecier", ok will switch to this :+1:

config.js
wonderingabout: also for naming i thought put indicator first to find the variables more easily in inspector (so we'll have 3 indicators in the indicator "family" one next to each other)

so maybe `indicatorDeprecation` ?

config.js
wonderingabout: i'm starting to understand it a little

so there are many things to notice, but the first thing i notice is you should put parameters in a function only if you want to process and modify that parameter

for example 

```
function f(x) {
    x = x*x;
    console.log(x);
}
```
is wrong, but then : 

```
function f(x) {
    return x*x;
}
```
will return the processed value, i see !!
very interesting :+1:

wonderingabout: so !!

if i am understanding it correctly, :1st_place_medal: 

then it means i should first convert the wanted arg to a universal one, then apply universal function for it, same as in the UHMAEATs

```
let universalArgument = "";
universalArgument = config.maxtotalgames;
deprecatedUniversalArgumentCheck();
universalArgument = config.maxtotalgames;
deprecatedUniversalArgumentCheck();
```

@windo i seee !!
very clear now !!

wonderingabout: just added OOP now :+1:

wonderingabout:
so,

will do a little break now
but big thanks @windo, i understand functions define a lot better now, see : https://github.com/online-go/gtp2ogs/pull/151#discussion_r

wonderingabout:
and actually i can use an array instead for faster code !!

config.js
wonderingabout: and actually i can use an array instead for faster code !!

config.js
wonderingabout: indeed !! :1st_place_medal:

config.js
wonderingabout: a bit complicated for me to understand it, as i dont know all the theory behind it :+1:

config.js
wonderingabout: in the future we may use this for `notification.user.color` or `notification.autohandicap` which can also have "automatic" as a value

config.js
wonderingabout: i see, so using an array then :1st_place_medal:

wonderingabout:
> Oh, and as for the failure on line 266 (`depreciatedArgumentUsedOne`), I think there is a difference in the name of the function argument and the name you are using to refer to it on that line?

for that i'll look at it later after i rework all this :+1:

wonderingabout:
big thanks for the thorough review, i learnt a lot :+1: @windo

config.js
windo: You can also create the alternatives part of the string explicitly, which might be better:

```
let msg = "Warning: --${deprecatedArg} is no longer supported. Use one of ";
let i = 0;
for (let e of alternatives) {
  if (i == 0) {
    msg += "--" + e;
  } else {
    msg += ", --" + e;
  }
msg += ".";
console.log(msg);
```

The `map()` followed by `join()` is just a shorter way of expressing the same:

`map(x => ("--" + x)` takes each element of an `Array` and (assuming it is a string) prepends it with a "--".

`join(", ")` takes all elements of an `Array` and joins them together into one string, with ", " separating each element.

Doing certain simple transformations on lists (`filter`, `map`, `reduce`) is a pretty common approach, see for example this blog post to illustrate:

https://medium.freecodecamp.org/ruby-using-the-select-map-and-reduce-methods-together-a9b2af30804b

But as you correctly point out, it can be harder to read the more complicated it gets. And maybe in this case it was already too much.

wonderingabout: i see, its much clearer now :+1: @windo 

but more importantly,
how can i extract the same from a variable

for example extract `--maxtotalgames` from `argv.maxtotalgames=20` is no longer supported

wonderingabout:
i'm writing a new code, if you want to start reviewing, still unfinished though : 
@windo 

```
    /*** console : DEPRECATIONS ***/
    console.log("-----DEPRECATIONS-----");
    let universalDeprecations = {
        toTestArray: [argv.botid, argv.bot, argv.id, argv.minrankedhandicap, argv.minunrankedhandicap, argv.maxrankedhandicap, argv.maxunrankedhandicap, argv.maxtotalgames, argv.maxactivegames],
        toUseArray: ["username", "username", "username", "minhandicapranked", "minhandicapunranked", "maxhandicapranked", "maxhandicapunranked", "maxconnectedgames", "maxconnectedgamesperuser"],
        indicator: false, // will become true if any deprecation is detected
        checkAndWarnOneArgument() {
            for (i=0 ; i < this.toTestArray.length ; i++) {
            /* always make sure toTestArray is always in sync with toUseArray 
            to avoid wronging the index number symetry, and to avoid breaking the loop by
            having one Array longer than the other array .length */
                if (this.toTestArray[i]) {
                    this.indicator = true;
                    console.log(`Warning: --${"toTestArray[i]"} is no longer supported. Use --${"toUseArray[i]"} instead.\nRead latest up to date available gtp2ogs arguments in OPTIONS-LIST (devel branch) for details : https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md`);
                }
            }
        },
        toTestOneArray: [argv.minmaintime, argv.minmaintimeranked, argv.minmaintimeunranked, argv.maxmaintime, argv.maxmaintimeranked, argv.maxmaintimeunranked, argv.minperiodtime, argv.minperiodtimeranked, argv.minperiodtimeunranked, argv.maxperiodtime, argv.maxperiodtimeranked, argv.maxperiodtimeunranked],
        toUseOneToThreeArray:
        
        
    };

universalDeprecations.checkAndWarnOneArgument();
```

i think this one is much better, thanks to all your feedback @windo :+1: 
but i still dont know how to extract the name from a variable and display the variable name, not its value

wonderingabout:
for easier review, continues here : 

https://github.com/online-go/gtp2ogs/pull/152

config.js
windo: Seems like a good try. "Inidcator deprecation" does not sound super-clear to me, if I see the name I think I would have to read a bit of code and see how it is used to understand what it means... An alternative might be `deprecationFound` maybe? But maybe that's just from codebases I've worked with.

I generally try to come up with the most clear but short name I can come up with and let reviewers suggest better names if the have any ideas. And if someone suggests something that's good then try to remember it :)

config.js
windo: I think you have to pass in `"maxtotalgames"` as a string, rather than the variable, that is:

```
checkDeprecation("maxtotalgames");
```

And not:

```
checkDeprecation(argv.maxtotalgames);
```

config.js
windo: Makes sense!

wonderingabout:
@windo 

thank you for the review
but this PR is an old version

new version (unfinished) is here : https://github.com/online-go/gtp2ogs/pull/152

PR #152, opened 2019-02-26
(unfinished) major console warnings rework v2

wonderingabout:
general look here : https://github.com/wonderingabout/gtp2ogs/tree/enhance-console-messages-display-v2/config.js

continuation of #151 
now using functions to avoid needless and unreliable
repetitions

also added an [OK!] or [ERROR(S)] display result

also pushing many fixes in in this commit as well

currently, code is being rewritten in a much smarter way thanks to very helpful review
for easier review, let's continue here :+1: 

@roy7 @Dorus @windo

### DIFF
--- a/config.js
+++ b/config.js
@@ -238,114 +238,147 @@ exports.updateFromArgv = function() {
         process.exit();
     }
 
-    // console : greeting //
-
+    // console : boot message //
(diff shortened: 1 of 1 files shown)


wonderingabout:
itworkslol

answer of my question, from : 

https://stackoverflow.com/questions/47468361/get-variable-name-into-string-in-javascript

roy7:
I do go ahead and merge #153 since it's a nice and clean approach.

PR #153, opened 2019-02-26
Moved deprecated messages to function.

Dorus:


Dorus:
This is my cleaner attemp at what @wonderingabout did in major console warnings rework v2 #152

Issue #27, opened 2017-03-14
Dependencies are apparently broken for npm installs


anoek:
Performing a `npm install -g gtp2ogs` should be all that is necessary, however something is wrong with dependency installation and thus the current work around is to `npm install optimistsocket.io-client` after they install gtp2ogs. This should not be necessary.

wonderingabout:
should be fixed in #157 

@anoek @roy7 @Dorus

roy7:
The thing was it should have always been working. But I will close this until it happens again.

wonderingabout:
@roy7 

i'm currently updating the gtp2ogs-tutorial while moving to ubuntu 18.04 at the same time

this has been tested to work

i also did `npm audit fix --force` which was also tested to work after that

should i send a new PR for that ?

![node202](https://user-images.githubusercontent.com/38690718/53682134-00d21c00-3cf3-11e9-9880-e42c9f3e6450.png)

wonderingabout:
tested to work ingame : 

![node203d](https://user-images.githubusercontent.com/38690718/53682143-15161900-3cf3-11e9-835a-c0352b40e06a.png)

PR #157, opened 2019-03-02
documentation updates for npm install from package.json

wonderingabout:
@anoek @roy7 @Dorus @windo 

`npm install` was indeed enough to install all needed
packages from the provided package.json, as @Dorus 
explained
(run as sudo for linux, or admin for windows)

in my test, eslint, mocha, and co, were all succesfully
detected and installed with npm install (need to cd
to gtp2ogs directory first)

also see screenshots included in the PR for examples

npminstall

npminstall2

in this example, i am upgrading from the installed gtp2ogs 5.3, 
so it is expected to see many packages installed

therefore, the optional steps to install optimist in case of error
are also not needed anymore : npm install will do that
too

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,14 +69,38 @@ This step is **is recommended**
 By default, npm installs an old branch of gtp2ogs that does not include latest 
 improvements, new features, and fixes
 
-To upgrade to devel branch (newest), see :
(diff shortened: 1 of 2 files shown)


wonderingabout:
this should fix #27

Issue #160, opened 2019-03-02
(bug) very long loop "Killing bot because of gamedata packet after bot was started" + bug 2 "We may need to move but were not able to. Re-connect to trigger action based on game state."


wonderingabout:
@roy7 @Dorus @windo 

using latest devel custom branch : https://github.com/wonderingabout/gtp2ogs/tree/roy7live-textonly-phoenixgo

got this very long loop during a game against royalminigo, is it normal ?

log too long, so pasted here : 
https://pastebin.com/iJrwJvjL

but basically it starts like this 

```
Mar 02 19:56:36   #  Max handicap is 0
Mar 02 19:56:36   #  Rejecting challenge from Littlemelon (2k)  [19x19] H6 id = 16817626
Mar 02 20:32:01   [Game 16818027]            Connecting to game.
Mar 02 20:32:01   [Game 16818027]            gamedata       B RoyalMinigo  [19x19]    
Mar 02 20:32:01   [Game 16818027]            Starting new bot process [24026]
Mar 02 20:32:01   [Game 16818027]            State loading for new bot
Mar 02 20:32:01   [Game 16818027]            genmove black
Mar 02 20:32:01 !!!!! [24026] stderr: 2019-03-02 20:32:01.983835: I model/zero_model.cc:72] Read checkpoint state succ
Mar 02 20:32:02 !!!!! [24026] stderr: 2019-03-02 20:32:02.001454: I model/zero_model.cc:80] Read meta graph succ
Mar 02 20:32:02   [Game 16818027]            gamedata       B RoyalMinigo  [19x19]    
Mar 02 20:32:02   [Game 16818027]            Killing bot because of gamedata packet after bot was started
Mar 02 20:32:02   [24026] Stopping bot
Mar 02 20:32:02   [Game 16818027]            Starting new bot process [24082]
Mar 02 20:32:02   [Game 16818027]            State loading for new bot
Mar 02 20:32:02   [Game 16818027]            genmove black
Mar 02 20:32:02 !!!!! [24082] stderr: 2019-03-02 20:32:02.121740: I model/zero_model.cc:72] Read checkpoint state succ
Mar 02 20:32:02 !!!!! [24082] stderr: 2019-03-02 20:32:02.142003: I model/zero_model.cc:80] Read meta graph succ
Mar 02 20:32:02   [24082] Stopping bot
Mar 02 20:32:02   [Game 16818027]            gamedata       B RoyalMinigo  [19x19]    
[...]

PR #162, opened 2019-03-03
Disconnect game in gameOver()

roy7:
Can this cause a race condition if active_game does arrive after gamedata, where we try to disconnect twice? If we have already disconnected, will the timeout here crash with an undefined error? Do we actually need more like

setTimeout(() => {  if (this && this.conn) this.conn.disconnectFromGame(gamedata.id);  }, 1000);

or similar? Or store the disconnecting timeout in the conn object and test for it so we don't set the timer a second time? Like a this.conn.shuttingdown.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -41,7 +41,6 @@ class Connection {
         this.connected_games = {};
         this.games_by_player = {};     // Keep track of connected games per player
         this.connected = false;
-        this.disconnect_timeout = null;
 
         this.connect_timeout = setTimeout(()=>{
             if (!this.connected) {
@@ -173,9 +172,10 @@ class Connection {
                 //     on server side: sometimes /gamedata event with game outcome is sent after
                 //     active_game, so it's lost since there's no game to handle it anymore...
                 //     Work around it with a timeout for now.
-                if (!this.disconnect_timeout) {
+                if (!this.connected_games[gamedata.id].disconnect_timeout) {
                     console.log("Starting disconnect Timeout in Connection active_game for " + gamedata.id);
-                    this.disconnect_timeout = setTimeout(() => {  this.disconnectFromGame(gamedata.id);  }, 1000);
+                    this.connected_games[gamedata.id].disconnect_timeout =
+                        setTimeout(() => {  this.disconnectFromGame(gamedata.id);  }, 1000);
                 }
             }
 
@@ -1648,7 +1648,6 @@ class Connection {
     }}}
     terminate() {{{
         clearTimeout(this.connect_timeout);
-        clearTimeout(this.disconnect_timeout);
         clearInterval(this.ping_interval);
         clearInterval(this.notification_connect_interval);
         clearInterval(this.corr_queue_interval);
--- a/connection.js
+++ b/connection.js
@@ -173,7 +173,10 @@ class Connection {
                 //     on server side: sometimes /gamedata event with game outcome is sent after
                 //     active_game, so it's lost since there's no game to handle it anymore...
                 //     Work around it with a timeout for now.
-                if (!this.disconnect_timeout) this.disconnect_timeout = setTimeout(() => {  this.disconnectFromGame(gamedata.id);  }, 1000);
+                if (!this.disconnect_timeout) {
+                    console.log("Starting disconnect Timeout in Connection active_game for " + gamedata.id);
+                    this.disconnect_timeout = setTimeout(() => {  this.disconnectFromGame(gamedata.id);  }, 1000);
+                }
             }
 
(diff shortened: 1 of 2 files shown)


wonderingabout:
continuation of #159

roy7:
Ok I added storing and testing for the disconnect timeout so it'll never be set more than once, just in case.

wonderingabout:
testing from now on, will tell you how it goes :+1:

wonderingabout:
testing looks good so far, see : 

https://github.com/online-go/gtp2ogs/issues/159#issuecomment-469178170

wonderingabout:
actually still error : 

https://github.com/online-go/gtp2ogs/issues/159#issuecomment-469212435

wonderingabout:
still error again v2 : 

stillerror2

roy7:
@wonderingabout Ok try again. It's possible the issue I found related to ignoring active_game finished events depending on whose turn it is was the problem. Or at least part of it.

wonderingabout:
ok, will try again right now @roy7 and tell you the result

i wish we remove apikey from debug though, and use username instead

wonderingabout:
done ;)

```
git remote add upstream https://github.com/roy7/gtp2ogs && git fetch upstream && git pull upstream gameover-disconnect
```

in https://github.com/wonderingabout/gtp2ogs/tree/roy7-test

wonderingabout:
will tell you how it goes @Dorus @roy7 @windo

wonderingabout:
ok, i'm testing it too atm, will tell you how it goes :1st_place_medal: 

and if it's fixed i'd like it if you hard reset force push all changes in one commit :+1: @roy7 

btw, you can send me games against my bot same as last time, i rarely have opponents at that time of the day

roy7:
When I merge I could use "squash and commit" or "rebase and commit" right on the web page here, never done that before though. Normally I merge as-is so the commit history isn't lost.

wonderingabout:
i may also try the squash thing, i generally just rebase manually, but if theres a faster and reliable way why not :+1:

connection.js
```diff
@@ -162,13 +162,6 @@ class Connection {
             if (gamedata.phase === "finished" && !(gamedata.id in this.connected_games))
```
windo: I wonder if it would be clearer if all handling for finished games was in one place:

```
// Handle finished games.
if (gamedata.phase === "finished") {
    if (!gamedata.id in this.connected_games) {
        return;
    } else {
        setTimeout(...);
    }
}

// Handle in-progress games.
...
```

wonderingabout: no idea at all :smile: 
but feedback always appreciated :100:

roy7:
Good idea, I'll redo this a bit tomorrow to make it cleaner.

wonderingabout:
@roy7 @Dorus @windo 

ok, i dont have the issue anymore now, looks good to be merged :1st_place_medal: 

consider committing cleanly would also be good too :+1: 
but now i'm getting the #160 issue often too, had it again this morning : 3 times in 3 days

connection.js
roy7: @windo I made the change, see what you think before I merge? I also made the logging based on DEBUG. Thought about removing it completely, but I guess it isn't bad to log that.

connection.js
```diff
@@ -158,24 +158,26 @@ class Connection {
...
+                    // XXX We want to disconnect right away here, but there's a game over race condition
+                    //     on server side: sometimes /gamedata event with game outcome is sent after
+                    //     active_game, so it's lost since there's no game to handle it anymore...
+                    //     Work around it with a timeout for now.
+                    if (!this.connected_games[gamedata.id].disconnect_timeout) {
+                        if (config.DEBUG) console.log("Starting disconnect Timeout in Connection active_game for " + gamedata.id);
+                        this.connected_games[gamedata.id].disconnect_timeout =
+                            setTimeout(() => {  this.disconnectFromGame(gamedata.id);  }, 1000);
+                    }
                 }
```
windo: I wonder if we should return here?

roy7: Good catch. In fact I was able to refine it further now. Take a look. :)

roy7:
I'm pretty sure this is good to go so I'll merge it. As a future TODO thought we could look at also calling gameOver( ) here as well, and then we could set up the disconnect timeout in gameOver() instead of two different places. Basically, if phase === finished, then gameOver(). We hopefully don't end up in situations with gameOver() called multiple times since we never reconnect to an already finished game any more.

wonderingabout:
@roy7 can you open a github issue as a reminder to yourself :+1:

wonderingabout:
and also close #159 too :+1:

PR #163, opened 2019-03-04
enhance console warnings with function checkThreeSameTimeFamily()

wonderingabout:
@roy7 @Dorus @windo 

makes managing console messages much easier and efficient

also moved function familyArrayFromGeneralArg(generalArg)
to the bottom of the code, same as other functions in other
.js classes of the code


wonderingabout:
tested to work today, it works !!!

PR #164, opened 2019-03-04
--min*max/periods/blitz*live*corr/+-ranked*unranked

wonderingabout:
@roy7 @Dorus @windo 

this PR is on top of #163 

didnt test it, but i hope it works
see commit message for explanations :+1: 

example of use : 

max time blitz : 30s + 30 x 5s
max time live : 2 hours + 20 x 2 minutes
max time corr : 7 days + 10x 3 days :+1:

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -158,9 +158,28 @@ class Connection {
                 this.processMove(gamedata);
             } */
 
-            // Don't connect to old finished games.
-            if (gamedata.phase === "finished" && !(gamedata.id in this.connected_games))
+            if (gamedata.phase === "finished") {
+                if (gamedata.id in this.connected_games) {
+                    // When a game ends, we don't get a "finished" active_game.phase. Probably since the game is no
+                    // longer active.(Update: We do get finished active_game events? Unclear why I added prior note.)
+                    // Note: active_game and gamedata events can arrive in either order.
+                    //
+                    if (config.DEBUG) conn_log(gamedata.id, "active_game phase === finished");
+
+                    // XXX We want to disconnect right away here, but there's a game over race condition
+                    //     on server side: sometimes /gamedata event with game outcome is sent after
+                    //     active_game, so it's lost since there's no game to handle it anymore...
+                    //     Work around it with a timeout for now.
+                    if (!this.connected_games[gamedata.id].disconnect_timeout) {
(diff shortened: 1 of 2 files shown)


wonderingabout:
@windo 

as for the existing options in https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md , the last renaming will be using plurals in #166 for `--bans` `--boardsizes` `--komis` `--speeds` `--timecontrols`

as for the new options i want to add until next master release, here is what i'm planning for now : 
- `--undo` and co
- `--colorseven` and co
- `--colorshandicap` and co

and maybe others will come up in my mind, but for now thats the main ones i want to add

wonderingabout:
renaming over and over can be indeed exhausting, but i chose to rename existing args to clearer ones for better futureproof, especially considering next master release is nearing

also, we added deprecation messages and warnings in case bot admin misses one, so hopefully it will ok

and then again, i didnt contribute for a long time in gtp2ogs so far, so these recent renamings are more like long term changes, i should not have any immediate nor other short term renaming plan, so i hope these will be the last for now :+1: @roy7 @Dorus @windo

wonderingabout:
and this renaming is not for being clearer, its for feature usability, as explained in commit message, now we can do for example : 

max time blitz : 30s + 30 x 5s
max time live : 2 hours + 20 x 2min
max time corr : 7 days + 10x 3 days :+1:

wonderingabout:
oh i forgot one last change i wanted to do, but would need some discussion : 

after @Dorus fully implements the config file in https://github.com/Dorus/gtp2ogs/tree/configFile , 
i wanted to remove all generalArgs and only keep the ranked and unranked argument for all family arguments, so that setting up gtp2ogs is much more intuitive (we would make it so that if ranked && !unranked then unranked=ranked, and vice versa, so no need to write settings twice), and this would also reduce complexity and number of argvs, so overall good
the main purpose of general arguments is that they save command line space, but if all settings are preset in a config file, they have little benefit i think, and the check to use the ranked value if unranked is unspecified and vice versa should remove the need for general arguments, but again we need to discuss this first :+1:

wonderingabout:
hmm i didnt close it weird

wonderingabout:
reopned as a duplicate in #173

PR #165, opened 2019-03-04
fix bypass min/max handicap limits if handicap is "automatic" (needs server support)

wonderingabout:
@anoek @roy7 @Dorus @windo 

update : this PR is now on top of latest devel, and now proposes `notification.bot.ranking` similar to `notification.user.ranking`
this PR addresses #28 
continuation of #121 

currently, there is an unwanted behaviour with
"automatic handicap" :

example : 

opponent 3d vs bot 7d
--minhandicap 0 --maxhandicap 2
automatic handicap -> a 7d-3d = 4 stones 
so a 4 stone handicap starts even though
--maxhandicap is set to 2

example 2 : 

opponent 3d vs bot 7d
--minhandicap 0 --maxhandicap 0
automatic handicap -> a 7d-3d = 4 stones 
so a 4 stone handicap starts even though
--maxhandicap is set to 0 
(no (0 max) handicap is bypassed when 
handicap is automatic)

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -470,20 +470,66 @@ class Connection {
         }
         /******** end of BOARDSIZES *********/
 
-        if (config.noautohandicap && notification.handicap === -1 && !config.noautohandicapranked && !config.noautohandicapunranked) {
-            conn_log("no autohandicap, rejecting challenge") ;
+        if (notification.handicap === -1 && config.noautohandicap) {
+            conn_log("no autohandicap");
             return { reject: true, msg: "For easier bot management, automatic handicap is disabled on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
 	}
 
-        if (config.noautohandicapranked && notification.handicap === -1 && notification.ranked) {
-            conn_log("no autohandicap for ranked games, rejecting challenge") ;
+        if (notification.handicap === -1 && config.noautohandicapranked && notification.ranked) {
+            conn_log("no autohandicap for ranked games");
             return { reject: true, msg: "For easier bot management, automatic handicap is disabled for ranked games on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
 	}
 
-        if (config.noautohandicapunranked && notification.handicap === -1 && !notification.ranked) {
-            conn_log("no autohandicap for unranked games, rejecting challenge") ;
+        if (notification.handicap === -1 && config.noautohandicapunranked && !notification.ranked) {
+            conn_log("no autohandicap for unranked games");
             return { reject: true, msg: "For easier bot management, automatic handicap is disabled for unranked games on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
 	}
+
+        /***** automatic handicap min/max handicap limits detection ******/
+
+        // below is a fix of automatic handicap bypass issue
+        // by manually calculating handicap stones number
+        // then calculate if it is within set min/max
+        // limits set by botadmin
+
+        // TODO : for all the code below, replace "fakerank" with 
+        // notification.bot.ranking (server support for bot ranking detection 
+        // in gtp2ogs)
+
+        if (notification.handicap === -1 && !config.noautohandicap && !config.noautohandicapranked && !config.noautohandicapunranked) {
+            let rankDifference = Math.abs(Math.trunc(user.ranking) - Math.trunc(config.fakerank));
+            // adding a trunk because a 5.9k (6k) vs 6.1k (7k) is 0.2 rank difference,
+            // but it is in fact a still a 6k vs 7k = Math.abs(6-7) = 1 rank difference game
+
+            // first, if ranked game, we eliminate > 9 rank difference
+            if (notification.ranked && (rankDifference > 9)) {
+                conn_log("Rank difference > 9 in a ranked game would be 10+ handicap stones, not allowed");
+                return {reject: true, msg: "Rank difference between you and this bot is " + rankDifference + "\n The difference is too big to play a ranked game with handicap (max is 9 handicap for ranked games), try unranked handicap or manually reduce the number of handicap stones in -custom handicap-"};
+            }
+
+            // then, after eliminating > 9 rank difference if ranked, we consider value of min-max handicap if set
+            // we eliminate all unwanted values, everything not forbidden is allowed
+            if (config.minhandicap && !config.minhandicapranked && !config.minhandicapunranked && (rankDifference < config.minhandicap)) {
+                automaticHandicapStoneDetectionReject("minhandicap");
+            }
+            if (config.minhandicapranked && notification.ranked && (rankDifference < config.minhandicapranked)) {
+                automaticHandicapStoneDetectionReject("minhandicapranked");
+            }
+            if (config.minhandicapunranked && !notification.ranked && (rankDifference < config.minhandicapunranked)) {
+                automaticHandicapStoneDetectionReject("minhandicap");
+            }
+            if (config.maxhandicap && !config.maxhandicapranked && !config.maxhandicapunranked && (rankDifference > config.maxhandicap)) {
+                automaticHandicapStoneDetectionReject("maxhandicap");
+            }
+            if (config.maxhandicapranked && notification.ranked && (rankDifference > config.maxhandicapranked)) {
+                automaticHandicapStoneDetectionReject("maxhandicapranked");
+            }
+            if (config.maxhandicapunranked && !notification.ranked && (rankDifference > config.maxhandicapunranked)) {
+                automaticHandicapStoneDetectionReject("maxhandicapunranked");
+            }
+        }
+        /***** end of automatic handicap min/max handicap limits detection ******/
+
 
         if (notification.handicap < config.minhandicap && !config.minhandicapranked && !config.minhandicapunranked) {
             minmaxHandicapFamilyReject("minhandicap");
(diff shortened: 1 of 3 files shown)


wonderingabout:
update : as explained in #167 ,
ideally, we would want to also have a new object `notification.handicap.automatic` that would tell us from server the actual number of stones when handicap is automatic (for example 3 stones, and not `-1`)

but as explained earlier, bot ranking detection would still be useful in other future commits

wonderingabout:
@anoek @roy7 
update : this PR is now on top of #170, and now proposes `notification.bot.ranking` similar to `notification.user.ranking`

no idea why travis still fails though

windo:
```
/home/travis/build/online-go/gtp2ogs/connection.js
  488:61  error  Irregular whitespace not allowed  no-irregular-whitespace
```

https://travis-ci.com/online-go/gtp2ogs/builds/103250953

wonderingabout:
yes thank you @windo, now i know how to look at travis thanks to @roy7 's explanations, but i didnt know what it meant lol

now that i know, i may take another look slower :+1:

wonderingabout:
can be downloaded here : 

https://github.com/wonderingabout/gtp2ogs/blob/fix-bypass-automatic-handicap-minmax-limits/gtp2ogs.js

only connection.js is changed here

windo:
It does say where the problem is:

```
/home/travis/build/online-go/gtp2ogs/connection.js
  488:61  error  Irregular whitespace not allowed  no-irregular-whitespace
```

It is in connection.js, on line 488, character 61 of the line, seems like here:

https://github.com/online-go/gtp2ogs/pull/165/files#diff-0d74c7455b8cbe5be6f644690ecb1b8aR488

wonderingabout:
but i dont see any problem here :sob:

windo:
Piping that line through `hexdump -C`

```
00000000  20 20 20 20 20 20 20 20  20 20 20 20 69 66 20 28  |            if (|
00000010  6e 6f 74 69 66 69 63 61  74 69 6f 6e 2e 72 61 6e  |notification.ran|
00000020  6b 65 64 20 26 26 20 28  72 61 6e 6b 44 69 66 66  |ked && (rankDiff|
00000030  65 72 65 6e 63 65 20 3e  20 39 29 29 c2 a0 7b 0a  |erence > 9))..{.|
```

You ca see that the last 3 bytes are `c2` `a0` `7b` and `0a`. So the space before `{` is not a normal space (`20`) but a "non-breaking space" (`&nbsp;` in html):

http://www.fileformat.info/info/unicode/char/00a0/index.htm

Just delete that space and press space normally to replace it and it should be all good.

wonderingabout:
delete a space, to replace it with a space.. :confused: :question: 

i dont mind doing it, but how did i insert a special space :confused:

wonderingabout:
it will be a bit dirty, but i'll commit on top of it to test if it works first (will rebase when all clear) :+1:

wonderingabout:
tested in https://github.com/wonderingabout/gtp2ogs/blob/fix-bypass-automatic-handicap-minmax-limits/connection.js#L488

wonderingabout:
whoaaaa
amazing, it works !! @windo

windo:
Cool!

I'm not sure what to think of this linter check. I mean, the explanation for the rule on [eslint page](https://eslint.org/docs/rules/no-irregular-whitespace) makes a little bit of sense, though I don't entirely understand what the problem with non-breaking space is - it did not seem to break nodejs for sure.

I think eslint itself is useful because I would forget about `===` if it didn't complain to me when I tried to use `==` and I would either get them back to the code base or a reviewer would have to spend their time pointing them out. And since I'm not a javascript expert, I can't tell which linter rules make sense and which ones do no not, so using the default ones as the starting point seems as good a choice as any - after all the people who have chosen the defaults probably had solid reasons for choosing them.

But if it keeps on causing problems (let's hope that this was an isolated case) to developers then that sucks...

wonderingabout:
no please keep it
it may have bugged me, but i learnt new things thanks to this

i think its a good thing too, we only need to get used to it :+1:

wonderingabout:
@anoek @roy7 @Dorus @windo 

just rebased with the no whitespace fix
and also rewrote the first PR message in a much clearer and shorter way

connection.js
```diff
@@ -459,6 +459,65 @@ class Connection {
         }
         /******** end of BOARDSIZES *********/
 
+        if (notification.handicap === -1 && config.noautohandicap) {
+            conn_log("no autohandicap");
+            return { reject: true, msg: "For easier bot management, automatic handicap is disabled on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
```
windo: This message implies handicap is allowed, but I suppose that's not guaranteed at this point?

So maybe just note that automatic handicap is disabled?

connection.js
```diff
@@ -459,6 +459,65 @@ class Connection {
...
+
+        if (notification.handicap === -1 && config.noautohandicapranked && notification.ranked) {
+            conn_log("no autohandicap for ranked games");
+            return { reject: true, msg: "For easier bot management, automatic handicap is disabled for ranked games on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
+	}
+
+        if (notification.handicap === -1 && config.noautohandicapunranked && !notification.ranked) {
+            conn_log("no autohandicap for unranked games");
+            return { reject: true, msg: "For easier bot management, automatic handicap is disabled for unranked games on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
+	}
```
windo: How about:

```
function checkRankedUnranked(option, notification, cb) {
  if (config[option]) {
    return cb("");
  }
  if (config[option+'ranked'] && notification.ranked) {
    return cb(" for ranked games");
  }
  if (config[option+'unranked'] && !notification.ranked) {
    return cb(" for unranked games");
  }
  return nil;
}

if (notification.handicap === -1) {
  let reject = checkRankedUnranked("noautohandicap", notification, (rur) => {
    conn_log("no auto handicap${rur}");
    return { reject: true, msg: "Automatic handicap is disabled${rur} on this bot" };
  });
  if (reject !== nil) {
    return reject;
  }
}
```

connection.js
```diff
@@ -459,6 +459,65 @@ class Connection {
...
+            }
+            if (config.maxhandicap && (rankDifference > config.maxhandicap)) {
+                conn_log("Automatic handicap " + rankDifference + " stones is too high, maximum handicap stone number is " + config.maxhandicap);
+                return {reject: true, msg: "Automatic handicap " + rankDifference + " stones is too high, maximum handicap stone number is " + config.maxhandicap + ", please reduce the number of handicap stones in -custom handicap-"};
+            }
+            if (config.maxhandicapranked && notification.ranked && (rankDifference > config.maxhandicapranked)) {
+                conn_log("Automatic handicap " + rankDifference + " stones is too high, maximum handicap stone number for ranked games is " + config.maxhandicapranked);
+                return {reject: true, msg: "Automatic handicap " + rankDifference + " stones is too high, maximum handicap stone number for ranked games is " + config.maxhandicapranked + ", please reduce the number of handicap stones in -custom handicap-"};
+            }
+            if (config.maxhandicapunranked && !notification.ranked && (rankDifference > config.maxhandicapunranked)) {
```
windo: I think you should figure out some way of reducing the repetition here.

connection.js
```diff
@@ -459,6 +459,65 @@ class Connection {
...
+            return { reject: true, msg: "For easier bot management, automatic handicap is disabled for unranked games on this bot, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones" };
+	}
+
+        /***** automatic handicap min/max handicap limits detection ******/
+
+        // below is a fix of automatic handicap bypass issue
+        // by manually calculating handicap stones number
+        // then calculate if it is within set min/max
+        // limits set by botadmin
+        if (notification.handicap === -1 && !config.noautohandicap && !config.noautohandicapranked && !config.noautohandicapunranked) {
```
windo: IIUIC, this will reject some handicaps that were automatically determined by OGS? I would not want to enable this for my bot because I'm happy letting the players play with whatever settings they like. So you probably should add an option to enable this behavior?

connection.js
wonderingabout: i see, then we can make it smarter : 

first we shorten the first message, then we add : 

if (config.maxhandicap === 0 && notification.handicap >= 1) "this bot does not play handicap games, please set handicap to -none-(0 stones)"

if config.minhandicap >= 1 && notification.handicap === 0) "this bot plays only handicap games, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones"

how about it @windo ? (would be done in a separate PR)

wonderingabout: update : i just realized we used the same "how about it" to end our message, when reading your 2nd message later :smiley:

connection.js
wonderingabout: the first line would need to be `if (config[option] && !config[option+"ranked"] && !config[option+"unranked"]) {`

this idea is interesting, but in some cases we need extra conditions, but yes we can start thinking about this and it would make code a lot more efficient

for the function name, more like `checkFamilyRejects` maybe

(and we dont need to pass into parameter `notification.` i think, isnt it a global variable ?)

and i feel like theres an easier way to write the reject function

and also would be done in a separate PR too @windo 

but yes definitely something i started to think about recently, and would be great to do :+1:

connection.js
wonderingabout: maybe we can start with this : https://github.com/online-go/gtp2ogs/pull/165#discussion_r263528457

connection.js
wonderingabout: no it will not @windo , if you look : 
- from line 483 : https://github.com/online-go/gtp2ogs/pull/165/files#diff-0d74c7455b8cbe5be6f644690ecb1b8aR483
- to line 491 : https://github.com/online-go/gtp2ogs/pull/165/files#diff-0d74c7455b8cbe5be6f644690ecb1b8aR491

the only thing that that code does is reject > 9 stones only if the game is ranked and automatic handicap is allowed : on ogs server, you cant play against a ranked game against someone 9 ranks stronger than you, so if automatic handicap picks more than 9 stones in a ranked game, it doesnt make sense at all : thats all this condition wants to do @windo 

then later in the code from line 493 https://github.com/online-go/gtp2ogs/pull/165/files#diff-0d74c7455b8cbe5be6f644690ecb1b8aR493 until the end, we only reject some specific values, and everything else is allowed

wonderingabout:
to sum up, the function for familyRejects would need to be done in another PR, way too many other places where it is needed to make it meaningful

but considering current code already works, its not my highest priority, i'd rather have all current PRs merged first @windo @roy7 @Dorus

wonderingabout:
@roy7 ty for adding wip, but the PR is really finished on my end

we need server support for either bot ranking detection, or sending us the number of handicap stones when handicap is automatic
can you look at that @anoek please ?
big thanks :+1:

roy7:
Yeah I marked it wip as a reminder to others not to merge it since server support is needed.

wonderingabout:
since server support may get some time to implement `notification.bot.ranking`, we may want to start to address it before master releaseby temporarily pushing a 
```--myfakerank 6d```
 kind of gtp2ogs argument, then replace it with `notification.bot.ranking` whenever server supports it

`--myfakerank` will use the same code as what is intended for `notification.bot.ranking`, so no conflict there

@anoek @roy7 @Dorus @windo

wonderingabout:
+ the way i see it, we dont need super precise rank estimation, only know whether rank is closer to 5d or 10k for example, just for handicap stones management, nothing more :+1:

roy7:
I wonder if we can't pull our ranking out of http://online-go.com/v1/me/settings/ as a work around.

wonderingabout:
i dont know about these pages @roy7 , but if theres any way as small as it is, i'll be very happy to implement it :100:

wonderingabout:
since it is using old code, reopening this in a new PR

PR #166, opened 2019-03-04
now using plurals for --bans --boardsizes --komis --speeds --timecontrols and co

wonderingabout:
@roy7 @Dorus @windo 

this PR is on top of #164 (now #173 )

see commit message for details :+1: 
didnt test it, i hope it works

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ Then, the following options are placed in the above ```<gtp2ogsarguments>```
 section. Put a space in between options when there are more than one.
 Also put a space in between the option and the parameter, for example :
 
-  ```--startupbuffer 2 --boardsize 13,19 --noclock --unrankedonly --maxconnectedgamesperuser 1 --maxmaintimelive 600 --maxperiodtimecorr 86400 --maxconnectedgames 10 --ban UserX,playerY --maxperiodsblitzranked 5 --maxperiodsliveranked 10 --maxperiodscorrranked 5```
+  ```--startupbuffer 2 --boardsizes 13,19 --noclock --unrankedonly --maxconnectedgamesperuser 1 --maxmaintimelive 600 --maxperiodtimecorr 86400 --maxconnectedgames 10 --bans UserX,playerY --maxperiodsblitzranked 5 --maxperiodsliveranked 10 --maxperiodscorrranked 5```
 
 **You can find a list of all possible to use gtp2ogs arguments here**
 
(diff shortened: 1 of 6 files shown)


roy7:
Hmm not sure how I feel about plurals. I think normally command line options that take "1 or more" arguments normally use the singular?

"All checks failed" btw. If you click Details next to the "continuous" line you can see the error messages.

wonderingabout:
@roy7 

this idea was suggested to me by @Dorus, and i think it makes sense, but maybe @Dorus can explain why

also, the fix in travis is easy 

```
config.allowed_sizes[19] = true;
```

need to be 

```
config.allowed_boardsizes[19] = true;
```

typo, will fix and check other remaining :+1:

wonderingabout:
actually i didnt do any mistake, its just that i didnt update test.js :smile: 

https://github.com/wonderingabout/gtp2ogs/blob/plurals-bans-boardsizes-komis-speeds-timecontrols/test/test.js#L18-L21

wonderingabout:
just did git merge from latest devel 

now the question is what do we want to do with this @roy7 @Dorus @windo 

didnt test it yet, but i hope it works (since the test.js includes a boardsizes 19 end to end game)

roy7:
Are you running this live on your own bot? There's soo many lines with renamed variables. :)

wonderingabout:
didnt test it yet @roy7 :smile: 

but i saw the test.js includes an end to end game test, so it should work

however, to make sure i'll test it first and tell you how it goes :+1:

roy7:
I appreciate it. I'm personally not a fan of just renaming things for its own sake all over the place but if you and @Dorus want it that's fine.

wonderingabout:
maybe we want to hear @Dorus 's thoughts on this again so we can have common agreement and understanding on this before making decision

to me, it looks natural to use plural since these plural families work the same way as arrays (they contain many values)

wonderingabout:
as for me its just an idea i had without much knowledge behind it

wonderingabout:
@roy7 

i'm very happy !!!

tested this : `--persist --noclock --komis 7.5,automatic --boardsizes 15,19 --speeds blitz,live --timecontrols byoyomi,fischer --minperiodsblitz 5 --minperiodslive 5 --minperiodscorr 5`

it all worked well at first try !! :smile: 
i'm so happy that now i am able to code things at first try without even testing them !!

thank you everyone for the review and always helpful advice !!!

@Dorus said on discord that he'll review this PR too, but it works as it is

also, maybe @windo has thoughts on it too ?

Dorus:
@roy7 Yeah i prefer proper names, and the reasoning here is that if we even want to rename these things, we better do it now before a mayor release with these params is made.

For the historical picture: I told meta in another Pr to use plural variable names for arrays. He extended that to a lot of other places, including command line params. Tbh i must say i kinda like these names.

wonderingabout:
@Dorus the only one that i find a bit weird is `--komisranked` for example, which may be read as "komi is ranked - is komi ranked"

but we'll get used to these eventually, and yes i also think its the good time to polish things before a major release

for same reasons, i plan to replace `let i of` with `let e of` after this is merged

then, `--colorseven` and `--colorshandicap` will use the same reasoning : "allowed" families get plural because they work like arrays :+1:

roy7:
@windo If you have no objection to this overhaul I'll get it merged. Thanks all.

wonderingabout:
thanks to you to @roy7 as always :1st_place_medal: :100: :1234: :+1:

roy7:
Ok merging. Lets hope nothing else breaks. ;)

wonderingabout:
thanks, this one was tested so i'm confident about it :+1:

Issue #127, opened 2019-02-12
add challenge player color detection and rejects options for even and handicap games


wonderingabout:
as said in #121 

last time a 9d was forced to give my bot 2 stones !

@anoek @roy7 @Dorus 

https://online-go.com/game/16356789

this is because default color is set to auto

but my bot, while being 6d, is superhuman already, so it can give 2 stones even to a pro player

forcing strong bots to rank up to 9d to avoid the issue is not efficient nor fast solution

so can we also add a way for bot admin to make his handicap bot to be color white (=> our bot will always give stones (white color) even if our opponent is higher ranked)

wonderingabout:
this is easily configurable if we know our bot's ranking : 

` if (rankDifference >=0 && ((notification.handicap == -1 ) || (notification.handicap > 0)) `

wonderingabout:
so we only reject game if opponent is not black color

we can guess the player's color indirectly from the settings in my example above

however if we have a `notification.user.color` (maybe it already exists ?), then its even easier (because theres the case where our opponent set color black, then we'll still reject game because we cant detect his color)

wonderingabout:
just made this issue clearer :+1:

wonderingabout:
imperfect but added `--challengercolors` in #230 

(it doesnt tell automatic color value (black/white) but it's a good start)

PR #167, opened 2019-03-05
add --colorseven --colorshandicap +/- ranked/unranked (needs server support)

wonderingabout:
@anoek @roy7 @Dorus @windo 

this PR is on top of #166 

as explained in #127 , sometimes in handicap games we end up 
giving handicap stones to weaker (but higher ranked) opponents, 
while our bot is superhuman (cant lose against humans even with 
handicap)
so in that example case, we would want to always give stones to 
bot opponents even if they are higher ranked, and reject game if 
that condition is not fullfilled

this PR addresses the issue expressed and proposes a generalization 
of challenge color rejects for 2 cases : 
- 1- even games
- 2- handicap games

for both cases, a default of allowing all colors is included, similar to how 
boardsizes, komis, speeds, and timecontrols families already work

to work, this PR needs server support to add `notification.color` 
to the challenge object
then we will be able to debug and test if the value of 
`notification.color` for "automatic" and "random" colors is 
`-1` or `null`, and slightly adjust this proposed code based 
on the result
in regards to that last point, this PR has a common point than with #165, 
where we want to know the automatic handicap stones true value 
(and not `-1` which is not a number of stones), except : 
- that here we should not have a bypass of min/max limits,
- and that we can start an initial support by just using an alias for "automatic" 
same as how notification.komi `null` is given the alias "automatic" in argv, 
so this is a mix of automatic handicap and automatic komi common issues, 
and then ideally we would like to have `notification.color.automatic` and 
`notification.color.random` too, so that for example is automatic or random 
color ends up chosing "black", we accept the game if `--colorshandicap black`

to sum up, we need server support for `notification.color` as a starter, 

big thanks :+1:

### DIFF
--- a/config.js
+++ b/config.js
@@ -39,6 +39,12 @@ exports.allow_all_komis_ranked = false;
 exports.allowed_komis_ranked = [];
 exports.allow_all_komis_unranked = false;
 exports.allowed_komis_unranked = [];
+exports.allowed_colors_even = {};
+exports.allowed_colors_even_ranked = {};
+exports.allowed_colors_even_unranked = {};
+exports.allowed_colors_handicap = {};
+exports.allowed_colors_handicap_ranked = {};
+exports.allowed_colors_handicap_unranked = {};
 exports.allowed_speeds = {};
 exports.allowed_speeds_ranked = {};
 exports.allowed_speeds_unranked = {};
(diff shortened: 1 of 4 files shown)


wonderingabout:
so to sum up, and ideally,

i hope we can add server support for (forgive if name is incorrect) : 
- `player.ranking (bot ranking detection)
- `notification.color` (challenge color detection)

and ideally : 
- `notification.handicap.automatic` (handicap stones number (0 to 36, , and not `-1`) when challenge handicap is "automatic")
- `notification.color.automatic` (player color ("black" or "white", and not `-1`) when challenge color is "automatic") 
- `notification.color.random`  (player color ("black" or "white", and not `null`) when challenge color is "random") 
- `komi.automatic` (actual komi value (any positive or negative number, and not `null`) when challenge color is "automatic") 

but we still want to keep the old `-1` and `null` values i think, for example allowing "automatic" komi is more powerful than allowing 5.5 if 9x9, 7.5 if 19x19, 0.5 if handi
but when we allow "automatic" komi (`null`) for example, then if the automatic komi has the same value than chosen komi, game is still rejected : for example in a 19x19 non handicap game, automatic komi will be set to 7.5, but if opponent chooses custom komi `7.5` then we'll get challenge rejected, but if he chooses komi "automatic" which will end up being 7.5 too, then game will be accepted, so less powerful currently

i think we want the best of both worlds, so keep current objects and variables as is, but add on top of that the earlier mentionned new properties in challenger object

@anoek @roy7 @Dorus @windo

wonderingabout:
just did git pull from latest devel :+1:

wonderingabout:
same comment as in https://github.com/online-go/gtp2ogs/pull/165#issuecomment-473623130

besides fixing merge conflicts later, we need server support for `notification.color` to advance on this PR @anoek :+1: 

big thanks

wonderingabout:
just did ```git pull upstream devel``` and merge commited, ready to merge but this needs server support for ```notification.color``` @anoek @Dorus @roy7 @windo

wonderingabout:
we should also make a function for the connection.js part of the code : will do when servers supports it :+1: 

for the part in config.js, a function is already ready, will submit it in the coming days i hope :+1:

Issue #158, opened 2019-03-02
add undo support (optionnally enablable with --undo --undoranked --undounranked)


wonderingabout:
@roy7 @Dorus @windo 

i noticed sometimes people misclick against my bot and they resign just after it because of undo impossible

i tested today against gnugo, royalleela and co, and undo is not possible

i noticed the code about undo is here : 

https://github.com/online-go/gtp2ogs/blob/devel/game.js#L34-L38

```
        // TODO: Command line options to allow undo?
        //
        this.socket.on('game/' + game_id + '/undo_requested', (undodata) => {
            this.log("Undo requested", JSON.stringify(undodata, null, 4));
        });
```

but my main motivation for adding undo support is teaching : 
it would be very instructive for players of all levels to roll back a lost game to try other combinations and see how the game would go !! :1st_place_medal: 
and i'll definitely advertise my bot as supporting undo in greeting when this is done :100: 

i'm thinking of adding support for undo optional with `--undo` `--undoranked` `--undounranked`
similar to what was done for `--showboard` in #143 (`--undounranked` seems the wisest one if you ask me, but i'll leave that to botadmin : maybe add a counter for undo `--maxundo` `--maxundoranked` `--maxundounranked` too ?)
also, it may break some bots, so default would be disabled, same as `--showboard`

but i dont know how to support undo

i can write all the documentation, describe and co, as always,
but i need help on the technical way to enable this, which i have no idea how :+1: 

big big thanks

windo:
I believe the tricksy aspects of an undo are:

  1. How to give time for the player to request an undo - if the bot responds in one second then the "request undo" button will already be gone
  2. How to deal with a `genmove` in flight when an undo has been requested - the bot is already making the next move

You may need to add an option to delay making moves (trigger `genmove` from a `setTimeout`?) and when an undo request is granted have to kill and reload the bot unless the bot also supports GTP `undo` (is that standard?)

wonderingabout:
there are indeed 2 aspects to the `undo` support

1- the easier part : connection support
[...]

PR #168, opened 2019-03-05
add --undo --undoranked --undounranked : initial support

wonderingabout:
@roy7 @Dorus @windo 

this PR is on top of #166 
this addresses #158 

coded the initial surface code for undo support
the technical undo code remains to be coded, 
but this is all i can do with my current skills, 
i hope it helps :+1: 

read commit message for details :+1:

### DIFF
--- a/game.js
+++ b/game.js
@@ -31,11 +31,44 @@ class Game {
 
         this.log("Connecting to game.");
 
-        // TODO: Command line options to allow undo?
-        //
-        this.socket.on('game/' + game_id + '/undo_requested', (undodata) => {
-            this.log("Undo requested", JSON.stringify(undodata, null, 4));
-        });
+        if (config.UNDO && !config.UNDORANKED && !config.UNDOUNRANKED) {
+            this.socket.on('game/' + game_id + '/undo_requested_in_any_type_of_game', (undodata) => {
+                this.log("Undo requested in any type of game", JSON.stringify(undodata, null, 4));
+                // allow the undo and process it ..., then
+                ////
+                ////
+                this.log("Undo request allowed and processed in any type of game, unlimited number of undos remaining", JSON.stringify(undodata, null, 4));
+            });
+        }
+
+        if (config.UNDORANKED && this.state.ranked) {
+            this.socket.on('game/' + game_id + '/undo_requested_in_ranked_game', (undodata) => {
+                this.log("Undo requested in ranked game", JSON.stringify(undodata, null, 4));
+                // allow the undo and process it ..., then
+                ////
+                ////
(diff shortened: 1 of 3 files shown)


game.js
```diff
@@ -31,11 +31,44 @@ class Game {
 
         this.log("Connecting to game.");
 
-        // TODO: Command line options to allow undo?
-        //
-        this.socket.on('game/' + game_id + '/undo_requested', (undodata) => {
-            this.log("Undo requested", JSON.stringify(undodata, null, 4));
-        });
+        if (config.UNDO && !config.UNDORANKED && !config.UNDOUNRANKED) {
+            this.socket.on('game/' + game_id + '/undo_requested_in_any_type_of_game', (undodata) => {
```
windo: I don't think you want to wrap this in `if (config.UNDO ...)` yet, the `if (...)` should probably be in the callback.

game.js
```diff
@@ -31,11 +31,44 @@ class Game {
...
+                this.log("Undo requested in any type of game", JSON.stringify(undodata, null, 4));
+                // allow the undo and process it ..., then
```
windo: This should be something like:

```
this.conn.emit('game/undo/accept', this.auth({
     game_id: this.game_id,
     move_number: undodata,
}));
```

wonderingabout: feel free to commit on top of my PR and submit it @windo, i'd be glad :+1: 

i dont know anything about the connection things, so i currently cant help any further :+1:

windo: I believe just this is all you need to get the bot to accept the undo. Then you'd also need to kill the bot I think. So maybe the `emit` from above, followed by an `this.ensureBotKilled()` and a maybe a `this.scheduleRetry()`. You might not need a retry technically because it would be the opponents move and we should get a new gamedata when they make a move anyway. Not sure.

But the PR should probably do something that is at least approximately capable of handling undos, not much reason to merge a configuration for something that is not supported, I think.

wonderingabout: i really cant do that easily, so if you or someone else `git pull upstream undo-undoranked-undounranked-v1` i think i would really like it

but whenever one wants/can of course

as for me, it's really all i can do on support, for now at least (with my current skills and understanding

wonderingabout:
just did git pull from latest devel :+1:

wonderingabout:
just pulled latest devel on this, now we can see only the undo specific changes commits

this work is still unfinished, and the rest of the undo support is beyond my current coding skills, 

but anyone is free to use it without needing to ask :+1:

wonderingabout:
this is not functional and unfinished
we are currently focused on other features, we can start again from #168 when we want to look at supporting undo again

PR #170, opened 2019-03-05
fixing user.ranking back to notification.user.ranking

wonderingabout:
@roy7 

this PR is on top of latest devel, not linked with any of the other PRs
this PR will be useful to force push #165 on top of it in a few minutes

in challenge rejects in connection.js,
user.ranking is somehow working, but it seems safer
to revert to the old 5.3 code notification.user.ranking
instead, as in :
https://github.com/wonderingabout/gtp2ogs/blob/release-5.3/gtp2ogs.js#L1369-L1379
and in latest devel :
https://github.com/online-go/gtp2ogs/blob/devel/connection.js#L1566

removing notification. from user.ranking is probably
a mistake i did, so reverting to default naming
for safety and robustness

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -314,43 +314,43 @@ class Connection {
             return { reject: true, msg: "Currently, " + Object.keys(this.connected_games).length + " games are being played by this bot, maximum is " + config.maxconnectedgames + " (if you see this message and you dont see any game on the bot profile page, it is because private game(s) are being played) , try again later " };
         }
 
-        if ((user.ranking < config.minrank) && !config.minrankranked && !config.minrankunranked) {
-            let humanReadableUserRank = rankToString(user.ranking);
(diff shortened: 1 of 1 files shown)


roy7:
`let user = notification.user;` is set at the top of the function and is used by other tests like user.username. It desn't seem this change is needed? Or else why not revert user.username as well and remove the user=notification.user at the start?

wonderingabout:
@roy7 you know better than me, i compared with 5.3 and found it weird, but i dont know the theory behind it,

i didnt see this `let user = notification.user;` code

so if the change happened sometime after 5.3, then theres no need to revert it indeed
and as for #165 , i can remove this commit too :+1:

roy7:
Yeah I think it is fine as is. :)

wonderingabout:
ok :smile:

PR #171, opened 2019-03-06
Neat up messages/behavior when handling bot failures and exits

windo:
Meant to address #160

It stops the bot restarts on un-changed gamedata and tries to neaten up some of the messages (and related behavior) when handling bot failures and exits.

### DIFF
--- a/bot.js
+++ b/bot.js
@@ -15,8 +15,11 @@ class Bot {
         this.command_callbacks = [];
         this.command_error_callbacks = [];
         this.firstmove = true;
-        this.ignore = false;   // Ignore output from bot ?
-        // Set to true when the bot process has died and needs to be restarted before it can be used again.
+        // Ignore output from bot, don't try to talk to it. Differs from "dead" below in that we don't know that the
+        // process has terminated yet.
+        this.unusable = false;
+        // Set to true when the bot process has died and needs to be restarted before it can be used again. Certainly no
+        // point in trying to talk to it.
         this.dead = false;
         // Set to true when there is a command failure or a bot failure and the game fail counter should be incremented.
         // After a few failures we stop retrying and resign the game.
@@ -26,7 +29,6 @@ class Bot {
             this.proc = child_process.spawn(cmd[0], cmd.slice(1));
         } catch (e) {
             this.log("Failed to start the bot: ", e);
-            this.ignore = true;
             this.dead = true;
             this.failed = true;
             return;
@@ -35,12 +37,12 @@ class Bot {
         if (config.DEBUG) this.log("Starting ", cmd.join(' '));
 
         this.proc.stderr.on('data', (data) => {
-            if (this.ignore)  return;
+            if (this.isUnusable()) return;
             this.error("stderr: " + data);
         });
         let stdout_buffer = "";
         this.proc.stdout.on('data', (data) => {
-            if (this.ignore)  return;
+            if (this.isUnusable()) return;
             stdout_buffer += data.toString();
 
             if (config.json) {
@@ -81,17 +83,21 @@ class Bot {
                         ++i;
                         this.log(lines[i]);
                     }
-                    this.failed = true;
                     this.command_callbacks.shift();
                     let eb = this.command_error_callbacks.shift();
-                    if (eb) eb(line.substr(1).trim());
+                    if (eb) {
+                        this.failed = true;
+                        eb(line.substr(1).trim());
+                    }
                 }
                 else {
                     this.log("Unexpected output: ", line);
-                    this.failed = true;
                     this.command_callbacks.shift();
                     let eb = this.command_error_callbacks.shift();
-                    if (eb) eb();
+                    if (eb) {
+                        this.failed = true;
+                        eb();
+                    }
                     //throw new Error("Unexpected output: " + line);
                 }
             }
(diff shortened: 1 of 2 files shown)


bot.js
```diff
@@ -332,7 +341,7 @@ class Bot {
     }
     
     loadState(state, cb, eb) { /* {{{ */
-        if (this.dead) {
+        if (this.isUnusable()) {
             if (config.DEBUG) { this.log("Attempting to load dead bot") }
```
roy7: Think it's okay here and below at 399 to still log 'dead' bot instead of 'unusable' bot?

bot.js
```diff
@@ -415,11 +424,13 @@ class Bot {
                 this.proc.stdin.write(str + "\r\n");
             }
         } catch (e) {
+            // I think this does not normally happen, the exception will usually be raised in the async write handler
+            // and delivered through an 'error' event.
             this.log("Failed to send command: ", str);
             this.log(e);
-            this.dead = true;
             this.failed = true;
```
roy7: Since we removed this.dead = true do we need to add this.unusable = true?

bot.js
```diff
@@ -470,14 +481,21 @@ class Bot {
...
+    }
+
     kill() { /* {{{ */
         this.log("Stopping bot");
-        this.ignore = true;  // Prevent race conditions / inconsistencies. Could be in the middle of genmove ...
-        this.dead = true;
-        this.command("quit");
+        if (!this.isUnusable()) {
+            this.command("quit");
+            this.unusable = true;
```
roy7: If command() is async should we set unusable before we send the quit command?

game.js
```diff
@@ -80,9 +80,13 @@ class Game {
             // check if we're missing a move and send it to bot out of gamedata. For now as a safe fallback just
             // restart the bot by killing it here if another gamedata comes in. There normally should only be one
             // before we process any moves, and makeMove() is where a new Bot is created.
-            //
-            if (this.bot) {
-                this.log("Killing bot because of gamedata packet after bot was started");
+            let gamedataChanged = (JSON.stringify(this.state) !== JSON.stringify(gamedata));
```
roy7: Nice trick. :)

roy7:
Has this fixed the issue so far or are you still seeing it in practice? Or is this issue/fix separate from the illegal move game crash problem?

windo:
It fixed the issue of crashloops. I don't think it "caused" the the illegal moves and the like, but it probably exposed another bug, at least with my config (and maybe local changes). So I had to add other fixes on top to really stop it from causing problems: I now also locally:

  1. retry uploadMove if it is not accepted - I think this is not needed because it only served to highlight I was trying to upload two moves in a row
  2. track whether a `makeMove` is in flight and refuse to start a new one when one is not complete - this might be a real bug because so far I haven't been able to relate to anything special I'm doing, but I have just spent ~hour looking into it so maybe I'm just not seeing something I've messed up that is in fact related to these code paths.

As your pull request fixes the issue, I think there is no hurry with this. I'll try to reformulate these changes to add extra logging and tracking and cleanups rather than a bugfix. I guess the `gamedataChanged` would be nice anyway to avoid the needless restarting of the bot?

roy7:
Ok I'll not merge this yet and you can adjust in future then we'll get it merged when ready.

It's definitely nice that if the gamedata hasn't changed then we don't need to restart things. :)

roy7:
@windo Just to confirm you want me to wait while you make more changes or should I merge this as-is and you can make more changes off of that?

windo:
Yes, please wait!

windo:
(I have some related updates locally that I need to clean up and split up into commits)

roy7:
Roger!

roy7:
Meta has been using this for 4-5 days now and his crash loops have totally stopped (https://github.com/online-go/gtp2ogs/issues/160). Perhaps we could merge only the part about seeing if gamedata has changed or not, and any other changes can come later if you are still working on it?

wonderingabout:
yes, so far i've been running this for 4-5 days and arround 100 games, and didnt experience any crashloop so far :+1: 

i cant be 100% sure that a crashloop wont come in say 1 or 2 weeks, but all i can say is it doesnt seem to break anything, so why not merge it and see :+1: 

i'm also fine continuing the testing though :+1:

roy7:
@windo Are you still kicking around out there? :)

roy7:
I wonder if my comment above "Since we removed this.dead = true do we need to add this.unusable = true?" is an issue that caused you to see an illegal move, as it appears that part of the code would no longer trigger isUnusable() but should?

The reason for the old this.ignore setting is I found if I was going to restart bot but an old bot delivered a move I hadn't processed yet, it'd still process the stdout of the old bot and make the move. Then my new bot also made a move and I tried to move twice. So if I know I'm killing bot I toggled ignore to sure I didn't listen any more.

I'm going to add the gamedata change check to a new PR and merge that now, since it fixes the loop issue.

roy7:
The loop specific code has been merged here https://github.com/online-go/gtp2ogs/pull/193

roy7:
Given it's age and part of the code has been merged already in other PRs, I'll go ahead and close this one for now. If it ever needs re-opened or you want to finish it up that's fine, just re-open. :) Thanks.

PR #174, opened 2019-03-12
rename let i of (index) with : let e of (element)

wonderingabout:
@roy7 @Dorus @windo 

this PR is on top of #166 

using a clearer and more correct syntax, as 
`let e of` refers to the element, not the index

didnt test it, but it should work :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ Then, the following options are placed in the above ```<gtp2ogsarguments>```
 section. Put a space in between options when there are more than one.
 Also put a space in between the option and the parameter, for example :
 
-  ```--startupbuffer 2 --boardsize 13,19 --noclock --unrankedonly --maxconnectedgamesperuser 1 --maxmaintimelive 600 --maxperiodtimecorr 86400 --maxconnectedgames 10 --ban UserX,playerY --maxperiodsblitzranked 5 --maxperiodsliveranked 10 --maxperiodscorrranked 5```
+  ```--startupbuffer 2 --boardsizes 13,19 --noclock --unrankedonly --maxconnectedgamesperuser 1 --maxmaintimelive 600 --maxperiodtimecorr 86400 --maxconnectedgames 10 --bans UserX,playerY --maxperiodsblitzranked 5 --maxperiodsliveranked 10 --maxperiodscorrranked 5```
 
 **You can find a list of all possible to use gtp2ogs arguments here**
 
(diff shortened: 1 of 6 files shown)


Dorus:
This is not right. `for (let i in arr)` does stand for index.
Try this in console:
```
for (let i in ['a', 'b']) { console.log(i) }
```
```
for (let e of ['a', 'b']) { console.log(e) }
```

afbeelding

game.js
```diff
@@ -398,8 +398,8 @@ class Game {
                 warnAndResign("place_free_handicap returned wrong number of handicap stones, resigning.");
                 return;
             }
-            for (let i in moves)                     // Sanity check
-                if (moves[i].pass || moves[i].x < 0) {
+            for (let e in moves)                     // Sanity check
+                if (moves[e].pass || moves[e].x < 0) {
```
Dorus: You would want to replace this with

```
for (let e of moves) 
    if (e.pass || e.x < 0) {
```
Else you if you keep it at `in`, it makes more sense to also keep `i` for index.

Dorus: Actually, i would use `move` instead of `e` to make it even more clear.

wonderingabout: @Dorus 

i didnt know, my bad !!

ok i'll revert it then :+1:

wonderingabout:
@Dorus just updated both PRs according to your comment, thanks ! :100: :1st_place_medal:

game.js
wonderingabout: i'm not sure move makes it clearer than i (index)

since we are referring to an index, i'd rather use `i` i think, 
but i dont know :+1:

Dorus: > i'm not sure move makes it clearer than i (index)

I literally gave you the code how to replace `let i in moves` with `let move of moves`.

Anyway not changing it is fine too, either way works. It's all just code cleanup after all.

wonderingabout: @Dorus 

i really dont know a lot about these things, i didnt want to ignore your advice, sorry :+1: 

but as i see things, `i` is indeed an index contrary to what i thought, so using `i` was actually fine, so i reverted to it

your code suggests not a renaming but a modification
it is indeed interesting to refer to `e` rather than `move[i]`

but since i dont know a lot about this part of the code, i think its better that i avoid modifying it
maybe in some more general code cleanup it will be meaningful to change this part as well as all other similar parts in the same commit :+1:

Dorus: Yes, if you change `i in` to `move of`, you change what the variable is. It goes from the index to the element. So instead of `moves[i]` later on to get the element, you can use `move` directly.

Anyway as i said, reverting is is perfectly fine too. I was just adding information for you to have a complete picture.

roy7:
Can you do a gil pull from devel, it seems to have all of the plural stuff showing in the diff when I look at 'changed files'. I don't know why it does that though when devel has that already imported.

wonderingabout:
thanks again :+1: 

yes i know github works a bit weird, it cant update with latest devel unless i merge commit, but these branches are really one on top of the other one without any change

you can click on the individual commit to see what each PR changes :+1:

PR #175, opened 2019-03-12
rename banned_ranked*unranked_users to banned_users_ranked*unranked + small readme update

wonderingabout:
@roy7 @Dorus @windo 

this PR is on top of #174 

minor consistency change, similar to how other allowed_ 
families work

hopefully we are nearing the end of all the renamings, i can 
understand that it can be too much, but these changes 
look good before a master release :+1: 

didnt test but it should work i hope :+1:

PR #176, opened 2019-03-15
Minmaxhandicap family rejects function, which also includes handicaponly and evenonly specific rejects

wonderingabout:
@roy7 @Dorus @windo 

this PR is on top of #175 

now uses an interesting function for easier code maintainance
and cleaner code

this may be the starting stone for similar functions before master relase

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,10 +174,11 @@ note 2 : if an argument has ranked and unranked in the same "family", use:
 - OR, if you want to specify different settings for ranked and unranked games, use 
 both the ranked and the unranked argument with wanted values, and then don't use the 
 general argument ! (see [notes H-](/docs/NOTES.md#h-) for details)
-- the exception that confirms this rule is `--bans` : since bans is a string, we allow 
-both general value `--bans A,B` AND specific values at the same time too 
-`--bansranked X,Y --bansunranked Z` , see [Options-List](/docs/OPTIONS-LIST.md) 
-for details
+- the exception that confirms this rule is `--bans` : since bans is a string that 
+does not conflict with its ranked/unranked options, it is possible to use both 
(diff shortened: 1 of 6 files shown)


wonderingabout:
just rebased this in a smarter and more functionnal way :+1: 

didnt test it, i hope it works :+1:

wonderingabout:
will rebase next time, tired for today lol, but it should work now

it was only an issue about the scope of `notification.handicap` being not defined in the original place where i put the function :+1:

wonderingabout:
just rebased this, ready to merge @Dorus @roy7 

(didnt test it though, but i hope it works) :+1:

roy7:
Maybe I just don't understand how it works.

There's a test like `if (!arg.includes("ranked"))` but we're calling it like `minmaxHandicapFamilyReject(config.minhandicap)` so 'arg' will be something like the number 4?

wonderingabout:
@roy7 i dont understand what you dont understand :smile: 

its true that the `if (!arg.includes("ranked"))` test is redundant, i thought of removing it, then chose to keep it for clarity, but i may remove it in a further PR (there are way too much PRs that depend on this one, would have to rebase a lot)

roy7:
The function calls need to be more like `minmaxHandicapFamilyReject( { "minhandicap": config.minhandicap } )` for this to do anything. @wonderingabout is working on fixing them. :)

wonderingabout:
@roy7 

what is fun is that i did it like that for other functions in other PRs, but will do it here too i guess :+1: 
ty for help :+1:

roy7:
Well if I merged broken stuff previously I hope we can relocate it and fix them.

wonderingabout:
yes, but since the other PRs have the same problem, better redo it cleanly :smile:

PR #178, opened 2019-03-16
Function minmaxperiodsblitzlivecorr family reject

wonderingabout:
@roy7 @Dorus @windo 

this PR is on top of #177 

very similar with #176 

didnt test it, but i hope it works :+1:

PR #179, opened 2019-03-16
Function genericallowedfamilies reject

wonderingabout:
@roy7 @Dorus @windo 

this PR is on top of #178 

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -424,21 +424,15 @@ class Connection {
 
         /* if square, check if square board size is allowed*/
         if (!config.allowed_boardsizes[notification.width] && !config.allow_all_boardsizes && !config.allow_custom_boardsizes && !config.boardsizesranked && !config.boardsizesunranked) {
-            let boardsizeSquareString = config.boardsizes;
-            conn_log("square board size " + notification.width + "x" + notification.height + " is not an allowed size");
-            return { reject: true, msg: "Board size " + notification.width + "x" + notification.height + " is not allowed, please choose one of these allowed board sizes " + boardsizeSquareToDisplayString(boardsizeSquareString)};
+            genericAllowedFamiliesReject("boardsizes", notification.width);
         }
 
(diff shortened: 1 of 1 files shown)


wonderingabout:
should be final version and work now :+1: 

(still didnt test it though)

wonderingabout:
as expected :smile:

wonderingabout:
update : this code was quite cool, but since i couldnt find how to easily extract name from an object property, adding an extra parameter to main function instead

bye, my old code below : 

```
function pluralAllowedFamilyArgToSingularString(toConvert) { /* {{{ */
    let converted = String(toConvert)                // for example String(config.speedsranked)
                    .split("config.")[1]             // for example : "speedsranked"
                    .split("unranked")[0]            // for example : "speedsranked"
                    .split("ranked")[0];             // for example : "speeds"
    converted = (Array.from(converted));             // for example ["s", "p", "e", "e", "d", "s"]
    converted.pop();                                 // for example ["s", "p", "e", "e", "d"]
    converted = converted.join("");                  // for example "speed"
    return converted;
} /* }}} */
```

wonderingabout:
@roy7 @Dorus @windo 

just rebased and fixed
didnt test the ogs code, but tested the function plural to singular to work :+1:

wonderingabout:
this test fail may be a good hint to stop using these aliases 

```
        let t = notification.time_control;
        let user = notification.user;
```

looking at gtp2ogs code, are used only in connection.js in 2 functions only, and all other parts of the code never use it

so it may make our code more solid

if its ok with you, i can push an all in one commit to remove it @Dorus @roy7 @windo @anoek

roy7:
The test fail was just missing a comma, nothing to do with the alias. :)

wonderingabout:
yes, ty @roy7 fixed now :+1: 

but now that we mentionned it, i still think it may be good to remove the `let user = notification.user` alias, but we can review that another time :+1:

PR #181, opened 2019-03-16
Function allowed family exports

wonderingabout:
@roy7 @Dorus @windo

this PR is on top of #180 

read commit message for details
didnt test, but i hope it works :+1:

PR #183, opened 2019-03-23
Fix bypass automatic handicap minmax limits using temporary --fakerank, and fix previously submitted functions, and add new functions as well

wonderingabout:
@anoek @roy7 @Dorus @windo 

rewrote #165 on top of #181 

while waiting for server support for bot ranking detection, 
using a reasonable approximation of bot ranking with
for example `--fakerank 6d`

examples of bypass issue : 

example :

opponent 3d vs bot 7d with --maxhandicap 2 :
automatic handicap -> a 7d-3d = 4 stones
so a 4 stone handicap starts even though
--maxhandicap is set to 2

example 2 :

opponent 1k vs bot 7d with
--minhandicap 2 --maxhandicap 5 :
automatic handicap -> a 7d-1k = 8 stones
so a 8 stone handicap starts even though
--maxhandicap is set to 8
(even (0 max) handicap is bypassed when
handicap is automatic)
the same behaviour also happens for minhandicap
as well
note : --minhandicap 0 disables automatic handicap
(because -1 < 0)

read commit message for details :+1: 

### DIFF
--- a/config.js
+++ b/config.js
@@ -577,11 +577,13 @@ exports.updateFromArgv = function() {
                     exports["allow_all_" + hashedArgNameStringConverted] = true;
                     // for example exports["allow_all_komis_ranked"] = true;
                 } else if (komi === "automatic") {
-                    exports[(prefixString + hashedArgNameStringConverted)[null]] = true;
-                    // for example exports["allowed_komis_ranked"[null]] = true;
+                    exports[prefixString + hashedArgNameStringConverted][null] = true;
+                    // for example exports["allowed_komis_ranked"][null] = true;
+                    // same as     exports.allowed_komis_ranked[null] = true;
                 } else {
-                    exports[(prefixString + hashedArgNameStringConverted)[komi]] = true;
-                    // for example exports["allowed_komis_ranked"[7.5]] = true;
+                    exports[prefixString + hashedArgNameStringConverted][komi] = true;
+                    // for example exports["allowed_komis_ranked"][7.5] = true;
+                    // same as     exports.allowed_komis_ranked[7.5] = true;
                 }
             }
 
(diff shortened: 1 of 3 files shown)


wonderingabout:
new problem now : 

we always reject even when we should not : 

problemnow

wonderingabout:
tested to work so far for boardsizes, maxhandicap, fakerank



worksyaay
fakerank
fakerank2

will try the other args later or another time :+1: 

i also want to push a sendchat for bot resigning after 5 crashes, but will do later i hope :+1:

roy7:
Glad we got that stuff fixed. Do you want it merged now or after you test more?

wonderingabout:
i'd like to finish the testing first :+1:

wonderingabout:
i think since everything else is working, we may just revert to old code for the allowed exports function, and fix it after master release

i'll try to do this as soon as i can :+1: 
so that we can release 6.0 in march 2019 :+1:

roy7:
I doubt we'll have a new master release this month, we need all the PRs to have some time to settle and be sure no one reports new errors. :)

Is this PR ready to merge then?

wonderingabout:
@anoek @roy7 @Dorus @windo @lemonsqueeze 

due to too short time, i reverted to old code for the allowed_ function

everything else was tested to work in connection.js and config.js too
i'd like if you all test it one last time, because i only reverted to old code, 
which was tested to work and be stable for many months now :+1:

wonderingabout:
i think when we all tested this one, we can release a new master release, and if there is anything we missed, release shortly after a fix like 6.0.1 or 6.1

but i'm confident everthing should work fine now :+1: 
(didnt test this latest commit though)

roy7:
Ok I'll merge it.

wonderingabout:
thank yoouu :smile:

PR #184, opened 2019-03-25
Hide API key in debug logging

roy7:


wonderingabout:
will definitely test this too after #183 is totally debugged !!

roy7:
Work as expected? :)

wonderingabout:
let me try :100:

wonderingabout:
so, maybe @Dorus or @windo or @anoek have an idea on this issue :100:

roy7:
@wonderingabout Did you ever figure out why this was failing?

wonderingabout:
@roy7 @Dorus 

no, but i was really looking forward to this

maybe our understanding of the `data` and `debug` code is not accurate :question:

wonderingabout:
maybe theres another way to look into this ? 
i think it would be a really nice improvement considering how apikeys appear on public messages recently :+1:

roy7:
I thought I was copying the object but I'm not, so I'm breaking the original object before we connect to the game. https://stackoverflow.com/questions/728360/how-do-i-correctly-clone-a-javascript-object

Dorus:
@roy7 To copy an object in modern javascript you can do `{ ...object }`

Do remember this will make a shallow copy, so if your objects contains sub items you should clonse those too.

Now the good part of this syntax is: You can make it not copy or replace certain elements at the same time, using this syntax:
`let noapidata = { ...data,  apikey: "hidden"};`

wonderingabout:
woaah !!!!

it works !! finally !!!

wooaaah

roy7:
@Dorus Think I should change it to your cleaner shallow copy version? I believe that requires Nodejs 8 to work when I saw it mentioned in the above threads. I ended up going with basic Stringify approach since it's super backwards compatible and performance doesn't matter (only done when a challenge arrives if we're debugging).

Dorus:
@roy7 Node 8.6 i believe, released 2017-09-26 and no longer supported. Current supported version is 8.16 and even that is just in maintenance, LTS is 10.x and current version is 12.x. Anybody running 8.5 or older should really upgrade.

Anyway you're right your version does a supper safe deep copy with good backwards compatibility and code isn't that ugly or anything. I would have used mine up front, and wouldn't hesitate to use it elsewhere, but i'm not sure if i would go back on the current already tested and working code.

wonderingabout:
personally i'm using the LTS version (currently nodejs 10) and it works fine :+1: 
(i also tested nodejs 11 flawlessly on ubuntu) :+1: 

for the code to test i dont mind doing other tests as long as you commit the tests and link me the github branches :smile: 

but seing as this is already good, maybe we can just merge it as it is, i didnt notice any performance issue with it :+1:

wonderingabout:
thanks :100: 

but i wonder why the tests failed

PR #191, opened 2019-05-06
fix minmaxrank typo

wonderingabout:
this PR is on top of latest devel 

fixing a minor typo introduced while converting old code to a function :+1: 

didnt test, but should work :100:


wonderingabout:
thanks :1st_place_medal:

PR #193, opened 2019-05-12
Fix restart loop by making sure gamedata has changed

roy7:
Doing same thing as https://github.com/online-go/gtp2ogs/pull/171 but leaving out the other changes @windo was still testing.

### DIFF
--- a/game.js
+++ b/game.js
@@ -82,8 +82,16 @@ class Game {
             // restart the bot by killing it here if another gamedata comes in. There normally should only be one
             // before we process any moves, and makeMove() is where a new Bot is created.
             //
-            if (this.bot) {
-                this.log("Killing bot because of gamedata packet after bot was started");
+            let gamedataChanged = (JSON.stringify(this.state) !== JSON.stringify(gamedata));
+
+            if (this.bot && gamedataChanged) {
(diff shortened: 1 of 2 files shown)


wonderingabout:
great ! :100:

roy7:
@Dorus From Discord discussion on recent gamedata packet problems, it came up that since

`this.state = gamedata;`

is before 

`const gamedataChanged = (JSON.stringify(this.state) !== JSON.stringify(gamedata));`

how does it ever fail this test?

game.js
```diff
@@ -82,8 +82,16 @@ class Game {
             // restart the bot by killing it here if another gamedata comes in. There normally should only be one
             // before we process any moves, and makeMove() is where a new Bot is created.
             //
-            if (this.bot) {
-                this.log("Killing bot because of gamedata packet after bot was started");
+            let gamedataChanged = (JSON.stringify(this.state) !== JSON.stringify(gamedata));
```
roy7, line 85: Does this need to be before line 47?

wonderingabout, line 85: could you use `const` instead of `let`

wonderingabout: edit: nevermind it seems this is not a current pr

PR #202, opened 2019-11-24
add rankedUnrankedGamesString function

wonderingabout:
added a function to avoid repeating 8 times
the same pattern for rankedunranked String

**still not tested** but should not be hard to fix
considering the code is relatively simple

will rebase in one clean commit when final version
is ready

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -338,15 +338,7 @@ class Connection {
         return { reject: false }; // OK !
 
         function minmaxRankFamilyReject(argNameString) {
-            // first, we define rankedUnranked, lowHigh, minMax, and humanReadableRank, depending on argNameString
-            let rankedUnranked = "";
-            // if argNameString does not include "ranked" or "unranked", we keep default value for rankedunranked
-            if (argNameString.includes("ranked") && !argNameString.includes("unranked")) {
-                rankedUnranked = "for ranked games ";
-            } else if (argNameString.includes("unranked")) {
-                rankedUnranked = "for unranked games ";
-            }
-
+            let rankedUnranked = rankedUnrankedGamesString(argNameString);
             let minMax = "";
             let lowHigh = "";
             if (argNameString.includes("min")) {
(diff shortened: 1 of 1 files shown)


connection.js
```diff
@@ -362,20 +354,12 @@ class Connection {
             let humanReadableMinmaxRank = rankToString(config[argNameString]);
 
             // then finally, the actual reject :
-            conn_log(`${user.username} ranking ${humanReadableUserRank} too ${lowHigh} ${rankedUnranked}: ${minMax} ${rankedUnranked}is ${humanReadableMinmaxRank}`);
-            return { reject: true, msg: `${minMax} rank ${rankedUnranked}is ${humanReadableMinmaxRank}, your rank is too ${lowHigh} ${rankedUnranked}` };
+            conn_log(`${user.username} ranking ${humanReadableUserRank} too ${lowHigh} ${rankedUnranked} : ${minMax} ${rankedUnranked} is ${humanReadableMinmaxRank}`);
```
Dorus: No need for spaces before `:`
It should read:
```
username ranking 2k too low for ranked games: min for ranked games is 1d.
```
not:
```
username ranking 2k too low for ranked games : min for ranked games is 1d.
```
Actually the whole sentence structure is a bit awkward, perhaps we can find a better format.

Also fancy idea: If a setting is the same for ranked/unranked, we could write
```
username ranking 2k too low to play: min rank to play is 1d.
```
bans already had that option. Again, we might want to find a sentence structure that's easy to build up.

connection.js
```diff
@@ -1369,6 +1302,35 @@ function pluralFamilyStringToSingularString(plural) { /* {{{ */
     return pluralToConvert;
 } /* }}} */
 
+function rankedUnrankedGamesString(argNameString) { /* {{{ */
+    // for the "bans" family :
+    if argNameString.includes("bans") {
```
Dorus: A pattern like this where a function contains one big if/else that both return is an indication it should have been 2 functions. Is it possible here to make a separate bans function and make the distinction between bans and others elsewhere?

You will also want to avoid to make the bans check everywhere else, but in places where you already know it'll always be bans or never be bans, you could call the right split function. If you end up having to do that bans check in multiple places, you could make 3 functions, one that does the ban check, and then calls the appropriate one of the other 2.

A quick scan trough the code shows we have `bannedFamilyReject`, that's probably the only one that needs to call the ban format string right?

Dorus: This should be 
```
if (argNameString.includes("bans")) {
```
right now this breaks the build.

wonderingabout: oh really ??
why is the () needed for a if ?
i thought it was not an obligation no ?

Dorus: Nvm this is already removed :)

connection.js
wonderingabout: fixed the no space needed

connection.js
wonderingabout: ok, then since the "bans" family is only one case, we'll just keep the rankedunranked specific code in the banned function
not worth creating one function that we'll use only in 1 case

wonderingabout: yes, i still dont understand if the problem was just an extra () after the if

wonderingabout:
just rebased cleanly in 1 commit, hopefully this works now (still needs testing though)

connection.js
Dorus: That's just the syntax for if statements in javscript. See https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Statements/if...else

> Syntax
```Js
if (condition)
   statement1
[else
   statement2]
```

Dorus: It's fine like this, but i like to itinerate that functions that do one small thing are good, even if they are only used once. For more background information, see this list: https://herbertograca.com/2016/09/03/clean-code-3-functions-by-robert-c-martin-uncle-bob/

wonderingabout:
i'll resubmit this idea in the #205 PR when testing is working, with clean commits

PR #203, opened 2019-11-24
Function minmax string list

wonderingabout:
this PR is on top of #202 

greatly improving old code with the new function
did not test it, but should work

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -338,15 +338,7 @@ class Connection {
         return { reject: false }; // OK !
 
         function minmaxRankFamilyReject(argNameString) {
-            // first, we define rankedUnranked, lowHigh, minMax, and humanReadableRank, depending on argNameString
-            let rankedUnranked = "";
-            // if argNameString does not include "ranked" or "unranked", we keep default value for rankedunranked
-            if (argNameString.includes("ranked") && !argNameString.includes("unranked")) {
-                rankedUnranked = "for ranked games ";
-            } else if (argNameString.includes("unranked")) {
-                rankedUnranked = "for unranked games ";
-            }
-
+            let rankedUnranked = rankedUnrankedGamesString(argNameString);
             let minMax = "";
             let lowHigh = "";
             if (argNameString.includes("min")) {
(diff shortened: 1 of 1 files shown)


connection.js
```diff
@@ -1369,6 +1278,27 @@ function pluralFamilyStringToSingularString(plural) { /* {{{ */
...
+        return "for unranked games";
+    } else {
+    // for example "boardsizes"
+        return "";
+    }
+} /* }}} */
+
+function minmaxStringList(argNameString) { /* {{{ */
+    if (argNameString.includes("min")) {
+        return ["minimum", "increase", "low"];
```
Dorus: You can add names to the elements by returning an object here instead of an array.

```Js
return ({min: "minimum", inc: "increase", low: "low"})
```
And the later call `minmax.min` instead of `minmaxList[0]}`.

connection.js
```diff
@@ -1020,42 +1001,26 @@ class Connection {
...
-                minMax = "Max";
-                increaseDecrease = "reduce";
-            }
+            let rankedUnranked = rankedUnrankedGamesString(argNameString);
+            let minmaxList = minmaxStringList(argNameString);
+            // ["minimum" or "maximum", "increase" or "decrease", "low" or "high"]
 
             // then, specific messages for handicaponly and evenonly messages first
-            if (notification.handicap === 0 && minMax === "Min" && config[argNameString] > 0) {
+            if (notification.handicap === 0 && minmaxList[0] === "minimum" && config[argNameString] > 0) {
```
Dorus: The check `minmaxList[0] === "minimum"` is terrible. Consider returning a boolean from the `minmaxStringList` function too. That way you can do `minmax.isMin`.

That would mean you need to extend the object i described below to `{min: "minimum", inc: "increase", low: "low", isMin: true}`.

Also fix this everywhere, i see this a bunch of times.

connection.js
wonderingabout: indeed !
i didnt know very well how to return a new object, it was really easy in fact lol

connection.js
wonderingabout: i see, indeed it is cleaner than if (A == true) !
thanks !

connection.js
```diff
@@ -1369,6 +1273,27 @@ function pluralFamilyStringToSingularString(plural) { /* {{{ */
...
+        return "for unranked games";
+    } else {
+    // for example "boardsizes"
+        return "";
+    }
+} /* }}} */
+
+function minMaxToStrings(argNameString) { /* {{{ */
+    if (argNameString.includes("min")) {
+        return {minMax: "minimum", increaseDecrease: "increase", lowHigh: "low", isMinMax: true};
```
Dorus: `isMinMax` is not the right name. `isMinMax` implies the boolean is true when it is min and/or max. You actually mean `isMin` is true and `isMax` is false. Since we only need to check one we can omit `isMax`.

Dorus: Actually i see the code did use `minMax === "Min"` and `minMax === "Max"`, so it could make sense to return both `isMin` and `isMax`. However you can also do `!minMax.isMin`, depending on your preferences.

connection.js
```diff
@@ -1369,6 +1273,27 @@ function pluralFamilyStringToSingularString(plural) { /* {{{ */
...
+    } else if (argNameString.includes("max")) {
+        return {minMax: "maximum", increaseDecrease: "decrease", lowHigh: "high", isMinMax: true};
```
Dorus: here you want to return `isMin: false` and/or `isMax: true`.

wonderingabout:
closing this as all changes are meaningless in further commits : 
- minmaxfunction is deprecated a few commits ahead
- rankedunranked function is not used alone

PR #204, opened 2019-11-24
Test umaeat

Dorus:
Added testing for error ban strings. This is on top of #203


Dorus:
@windo If you have time, can you give your opinion on these tests? Did i set them up correctly?

I will try to get around and update the PR once meta is done with all his refactoring pr's.

PR #206, opened 2019-11-27
Pass all notifications as parameters, move functions to bottom of the code, code cleaning and optimizations

wonderingabout:
on top of #207 

tested to work, ready to merge
note : now we shrinked connection.js below the 1000 lines of code threshold :1st_place_medal: ,
as compared to 1470 lines last week

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -39,7 +39,7 @@ class Connection {
         });
 
         this.connected_games = {};
-        this.games_by_player = {};     // Keep track of connected games per player
+        this.games_by_player = {};  // Keep track of connected games per player
         this.connected = false;
 
         this.connect_timeout = setTimeout(()=>{
@@ -88,13 +88,11 @@ class Connection {
 
         if (config.corrqueue) {
             // Check every so often if we have correspondence games that need moves
-            //
             this.corr_queue_interval = setInterval(() => {
                 // If a game needs a move and we aren't already working on one, make a move
-                //
                 if (Game.corr_moves_processing === 0) {
-                    // Choose a corr game to make a move
-                    // TODO: Choose the game with least time remaining
+                    /* Choose a corr game to make a move
+                    /  TODO: Choose the game with least time remaining*/
                     let candidates = [];
                     for (let game_id in this.connected_games) {
                         if (this.connected_games[game_id].corr_move_pending) {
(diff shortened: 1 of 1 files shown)


wonderingabout:
tested to work, ready to merge now :1st_place_medal: 
test j6
test j5
test j4
test j3
test j2
test j1

connection.js
```diff
@@ -313,7 +309,7 @@ class Connection {
             return { reject: true, msg: "Currently, " + Object.keys(this.connected_games).length + " games are being played by this bot, maximum is " + config.maxconnectedgames + " (if you see this message and you dont see any game on the bot profile page, it is because private game(s) are being played) , try again later " };
         }
 
-        const resultRank = minmaxHandicapRankActualReject("rank", user.ranking, notification.handicap, notification.ranked);
+        const resultRank = minmaxHandicapRankActualReject("rank", notification.user.ranking, notification.handicap, notification.ranked);
```
Dorus: This seems to be the online line in the entire function that uses something other than `notification.user`, If you move `minmaxHandicapRankActualReject` outside this function, you can replace the `notification` parameter with the `user` paramter. So `checkUser(notification)` -> `checkUser(user)` and you then call `checkUser(notification.user)`.

connection.js
```diff
@@ -920,56 +819,74 @@ function pluralFamilyStringToSingularString(plural) { /* {{{ */
...
+        return `${allowedBannedSetting}_unranked`;
+    } else if (argNameString.includes("ranked")) {
+        return `${allowedBannedSetting}_ranked`;
     } else {
-        return familyArray[0];
+        return allowedBannedSetting;
+    }
+} /* {{{ */
+
+function checkAndGenerateArgNameStringObject(familyObjectArgNameStrings, notificationRanked) { /* {{{ */
```
Dorus: The return type here is a string, not a stringObject.

I'm used to languages that have types build in, but since Js doesn't do that you should at least keep naming consistent with the type returned.

Another strategy is to name things not after their type but what they are. So for example use plural form to indicate an array instead of adding `array` suffix. But again that might be more suited for typed languages.

wonderingabout: fixed in https://github.com/online-go/gtp2ogs/pull/209/commits/e9e6b222a8a595cea0ca3c61ff009343464bea9a
thanks ! :1st_place_medal:

connection.js
wonderingabout: opted for a different idea in that approaches this problem more globally, in : https://github.com/online-go/gtp2ogs/pull/209/commits/d3365737df82a1255ba6e7fb5366e42c6330165a
thanks :+1:

PR #205, opened 2019-11-24
UHMAEAT 4.0 (main/period)times

wonderingabout:
major optimizations and simplifications over the old code

edit : you can also see the changes here directly : 
https://github.com/wonderingabout/gtp2ogs/blob/UHMAEAT-4-0/connection.js#L651

edit 2 : now tested and ready to merge :+1:

PR #207, opened 2019-11-29
Function minmax v2 periods / rank / handicap (notification.handicap) / handicap (fakerank) 

wonderingabout:
on top of #205 

tested and ready to merge now

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -316,55 +316,12 @@ class Connection {
             return { reject: true, msg: "Currently, " + Object.keys(this.connected_games).length + " games are being played by this bot, maximum is " + config.maxconnectedgames + " (if you see this message and you dont see any game on the bot profile page, it is because private game(s) are being played) , try again later " };
         }
 
-        if ((user.ranking < config.minrank) && !config.minrankranked && !config.minrankunranked) {
-            return minmaxRankFamilyReject("minrank");
-        }
-        if ((user.ranking < config.minrankranked) && notification.ranked) {
-            return minmaxRankFamilyReject("minrankranked");
-        }
-        if ((user.ranking < config.minrankunranked) && !notification.ranked) {
-            return minmaxRankFamilyReject("minrankunranked");
-        }
-        if ((user.ranking > config.maxrank) && !config.maxrankranked && !config.maxrankunranked) {
-            return minmaxRankFamilyReject("maxrank")
-        }
-        if ((user.ranking > config.maxrankranked) && notification.ranked) {
-            return minmaxRankFamilyReject("maxrankranked");
-        }
-        if ((user.ranking > config.maxrankunranked) && !notification.ranked) {
-            return minmaxRankFamilyReject("maxrankunranked");
+        const resultRank = minmaxHandicapRankActualReject("rank", user.ranking, notification.handicap, notification.ranked, user.username);
+        if (resultRank) {
+            return (resultRank);
         }
 
         return { reject: false }; // OK !
-
-        function minmaxRankFamilyReject(argNameString) {
-            // first, we define rankedUnranked, lowHigh, minMax, and humanReadableRank, depending on argNameString
-            let rankedUnranked = "";
-            // if argNameString does not include "ranked" or "unranked", we keep default value for rankedunranked
-            if (argNameString.includes("ranked") && !argNameString.includes("unranked")) {
-                rankedUnranked = "for ranked games ";
-            } else if (argNameString.includes("unranked")) {
-                rankedUnranked = "for unranked games ";
-            }
-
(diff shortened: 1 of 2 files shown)


connection.js
```diff
@@ -316,70 +316,20 @@ class Connection {
...
-            return minmaxRankFamilyReject("maxrank")
-        }
-        if ((user.ranking > config.maxrankranked) && notification.ranked) {
-            return minmaxRankFamilyReject("maxrankranked");
-        }
-        if ((user.ranking > config.maxrankunranked) && !notification.ranked) {
-            return minmaxRankFamilyReject("maxrankunranked");
+        const resultRank = genericMinmaxFamiliesReject("rank", user.ranking, notification.time_control.speed, notification.ranked, notification.handicap, user.username);
+        if (resultRank) {
+            return (resultRank);
```
Dorus: I'm probably at fault for introducing this style, but no need for those brackets here. `return resultRank;` will work too.

config.js
```diff
@@ -592,5 +562,5 @@ exports.updateFromArgv = function() {
             process.exit();
         }
     }
-
+
```
roy7: Bit of stray whitespace? :)

connection.js
```diff
@@ -571,653 +492,172 @@ class Connection {
...
+        function UHMAEAT(mainPeriodTime) {
+            for (let blitzLiveCorr of ["blitz", "live", "corr"]) {
+                if (t.speed === convertBlitzLiveCorr(blitzLiveCorr)) {
+                    const minFamilyArray = familyArrayMIBLIsminIsmaxFromGeneralArgString("min" + mainPeriodTime + blitzLiveCorr);
+                    const maxFamilyArray = familyArrayMIBLIsminIsmaxFromGeneralArgString("max" + mainPeriodTime + blitzLiveCorr);
+                    // example : ["min(main/period)timeblitz", "min(main/period)timeblitzranked", "min(main/period)timeblitzunranked", 
+                    //            ["minimum", "increase", "below", "low"], [true, false]]
+
+                    minmaxRejectTests(minFamilyArray);
+                    minmaxRejectTests(maxFamilyArray);
```
Dorus: Shouldnt the values retuned from these functions be returned somehow? Possibly only if they are not `null`?

Small Js trick that could work is `return minmaxRejectTests(minFamilyArray) || minmaxRejectTests(maxFamilyArray);`

connection.js
```diff
@@ -571,653 +492,172 @@ class Connection {
...
+                }
+            }
+
+            function minmaxRejectTests(familyArray) {
```
Dorus: Consider replacing `familyArray` with `[all, ranked, unranked]` and then replace `familyArray[0]` with `all`, `familyArray[1]` with `ranked` and `familyArray[2]` with `unranked`.

```
function minmaxRejectTests([all, unranked, ranked]) {
  if (config[all] || config[unranked] || config[ranked]) {
```
Another option is
```
function minmaxRejectTests(familyArray) {
  if (familyArray.some(fam => config[fam])) {
```

wonderingabout:
changes to config.js have been moved separately to #208

wonderingabout:
it seems current fail comes from generic minmax function, because this change (though incomplete and still mistaken) makes the test pass

https://github.com/online-go/gtp2ogs/pull/207/commits/53b39e9366042f855974f5f40a3a349785e15d8d

wonderingabout:
actually the more i look at it the more it seems needlessly complicated, so i'll just create 2 different functions 
1) minmaxperiods
2) minmaxrank, minmaxhandicap, minmaxhandicap(fakerank)

wonderingabout:
the mistake was small, but now i want to separate it in 2 other functions, seems wiser

wonderingabout:
finished adding many tweaks and fixes
these simply the reject algorithm (quick exit if we wont reject), remove repetitions (mistakes easy to spot and fix), and ads tweaks and fixes here and there

wonderingabout:
currently testing, most cases work, but currently i'm having an issue with the if all or ranked or unranked pattern :

test h0

wonderingabout:
fixed typo in test : minhandicap = 2, maxhandicap = 5

connection.js
wonderingabout: maybe you're right !
but since i dont know about this i cant tell what should be done except from the safe simple way

wonderingabout:
reminder for myself : for canadian periodtime test if `argNumber * notificationT.stones_per_period` or `notificationT.period_time` (which is already for X stones)

wonderingabout:
tested and should be ready to merge now :)
testi3 fakerank 6d
test i3
testi2
testi1
test h11

PR #209, opened 2019-12-06
checkChallenge PreRequirements/Settings

wonderingabout:
on top of #206 
tested to work, ready to merge and compatible with both #214 and devel

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -277,60 +277,56 @@ class Connection {
 
 
     } /* }}} */
-    // Check user is acceptable
+    // Check challenge pre-requirements are acceptable, else no need to check challenge settings
     //
-    checkUser(notification) { /* {{{ */
-        let user = notification.user;
+    checkChallengePreRequirements(notification) { /* {{{ */
 
-        if (config.banned_users[user.username] || config.banned_users[user.id]) {
-            return bannedFamilyReject("bans", user.id, user.username);
-        } else if (notification.ranked && (config.banned_users_ranked[user.username] || config.banned_users_ranked[user.id])) {
-            return bannedFamilyReject("bansranked", user.id, user.username);
-        } else if (!notification.ranked && (config.banned_users_unranked[user.username] || config.banned_users_unranked[user.id])) {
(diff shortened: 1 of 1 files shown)


connection.js
```diff
@@ -843,7 +852,7 @@ function familyObjectMIBLIsminmax(familyNameString) { /* {{{ */
         ba = "above";
         lh = "high";
     }
-    const familyArray = ["", "ranked", "unranked"].map(e => familyNameString + e);
+    const familyArray = familyArrayFromFamilyNameString(familyNameString);
     return {argNameStrings: {all: familyArray[0], ranked: familyArray[1], unranked: familyArray[2]},
```
Dorus: Why not make `familyArrayFromFamilyNameString` return the  `{all: familyArray[0], ranked: familyArray[1], unranked: familyArray[2]}` object?

Dorus: Then you can do `return { argNameStrings: familyArrayFromFamilyNameString(familyNameString) };`

Dorus:
```
function familyArrayFromFamilyNameString(familyNameString) {
    return {
        all: familyNameString,
        ranked: familyNameString + "ranked",
        uranked: familyNameString + "unranked"
      };
}
```

connection.js
```diff
@@ -485,35 +494,31 @@ class Connection {
...
+        if (c2.reject)  return c2;
 
         return { reject: false };  /* All good. */
 
     } /* }}} */
     on_challenge(notification) { /* {{{ */
-        let c = this.checkChallenge(notification);
-        let rejectmsg = (c.msg ? c.msg : "");
+        const c0 = this.checkChallenge(notification);
+        const rejectmsg = (c0.msg ? c0.msg : "");
```
Dorus: This can also be written as
```
const rejectmsg = c0.msg || "";
```

Dorus:
From what i can see, this PR looks good. 90% of it is just renaming/moving/refactoring code. Only functional change is the change in the order the checks are done.

PR #212, opened 2019-12-08
remove repetitions for nopause check in game.js, add a sendchat

wonderingabout:
on top of devel
optimizing an old pattern i introduced in the past
didnt test it, but should work as is


wonderingabout:
will be addressed in #213 directly instead

PR #221, opened 2020-03-29
(wip) exports ranked unranked

wonderingabout:
trying to split pr213 in smaller chunks

this PR is on top of #231 

reminder: when #230 and #231 are merged, incorporate some must-add changes from #232 before merging this PR , including those of the `--bans` code


connection.js
```diff
@@ -682,282 +626,142 @@ function conn_log() {
...
-    if (R >= 30)  return (R-30+1) + 'd'; // R>=30 : 1 dan or stronger
-    else          return (30-R) + 'k';   // R<30  : 1 kyu or weaker
-}
-
-function bannedFamilyReject(argNameString, uid, notificationUid) {
-    const rankedUnranked = beforeRankedUnrankedGamesSpecial("from ", "", argNameString, "all ");
-    conn_log(`${uid} ${notificationUid} is banned ${rankedUnranked}`);
-    return { reject: true, msg: `You (${uid} ${notificationUid}) are banned ${rankedUnranked} on this bot by bot admin, you may try changing the ranked/unranked setting` };
+function checkUserIsInSet(user, currSet) {
+    if (currSet.size === 0) return false;
```
Dorus: There is no need for this check as the line below will always be false for an empty set.

connection.js
```diff
@@ -269,45 +269,23 @@ class Connection {
...
-            }
-            if (!notification.ranked && config.banned_users_unranked[notification.user[uid]]) {
-                return bannedFamilyReject("bansunranked", uid, [notification.user[uid]]);
-            }
+        if (checkUserIsInSet(notification.user, config_r_u["blacklist"])) {
+            return getBlackWhiteListReject("blacklist", r_u_strings.for_r_u_games);
         }
-        const resultRank = minMaxHandicapRankRejectResult("rank", notification.user.ranking, false, notification.ranked);
-        if (resultRank) return resultRank;
+        if (!checkUserIsInSet(notification.user, config_r_u["whitelist"])) {
```
Dorus: Here we should also check if the set contains items, if it contains no items then bot admin is not using the whitelist.

wonderingabout: oh actually if whitelist is empty then the check returns false, so !check is true, so we'd always reject !!
i wanted to avoid a repetition but changed the meaning of the code !
thanks, fixed !

connection.js
wonderingabout: indeed, fixed !!!

connection.js
```diff
@@ -682,282 +626,141 @@ function conn_log() {
...
 }
 
-function noAutohandicapReject(argNameString) {
-    const rankedUnranked = beforeRankedUnrankedGamesSpecial("for ", "", argNameString, "");
-    conn_log(`no autohandicap ${rankedUnranked}`);
-    return { reject: true, msg: `For easier bot management, -automatic- handicap is disabled on this bot ${rankedUnranked}, please manually select the number of handicap stones you want in -custom handicap-, for example 2 handicap stones, you may try changing the ranked/unranked setting` };
+function getBlackWhiteListReject(setName, for_r_u_games) {
+    conn_log(`user is not allowed ${for_r_u_games} (${setName})`);
+    return { reject: true, msg: `This bot uses a ${setName}: your user name/id `
+                                + `is not allowed on this bot ${for_r_u_games}, `
```
Dorus: I would prefer "This bot uses a ${setName}: You are not allowed on this bot ${for_r_u_games}". mentioning `name/id` gives out unnecessarily info to malicious users, and plain confusing info for others.

connection.js
```diff
@@ -325,50 +322,35 @@ class Connection {
...
-
-        // TODO : add all sanity checks here of all unhandled notifications
-        // Sanity check, user can't choose rules. Bots only play chinese.
-        if (!notification.rules.includes("chinese")) {
-            conn_log(`Unhandled rules: ${notification.rules}`);
-            const msg = `The ${notification.rules} rules are not allowed for `
-                        + `this bot, please choose allowed rules such as `
-                        + `chinese rules. `;
-            return { reject: true, msg };
-        }
```
Dorus, line 338: Why did we remove this check? Ogs today still enforces Chinese rules for bots right?

config.js
```diff
@@ -248,213 +229,164 @@ exports.updateFromArgv = function() {
                 + `please visit https://github.com/online-go/gtp2ogs/tree/devel`
                 + `\nDebug status: ${debugStatus}`);
     // B - check deprecated argv //
-    testDeprecatedArgv(optimist.argv, "komis");
+    testDeprecatedArgv(argv, "komis");
```
Dorus, line 232: What is the use for passing a hardcoded "komis" parameter? Can't we inline this in the function?

wonderingabout:
this pr will never be merged, needs to be chunked in a lot of smaller ones, which we are going to do

PR #222, opened 2020-04-01
various refactoring from PR #213

wonderingabout:
on top of latest devel, incorported easy to review and minor changes from PR #213

one significant change is the deprecation of `sprintf` in package.json, replaced with template litterals

see TODO list for next PRs in commit notes

not tested yet but should work as is or with minor fixes as it doesn't modify the old code behaviour

PR #223, opened 2020-04-01
remove duplicates black/white in loadClock()

wonderingabout:
this is meant to be merged after PR #222 , so that we don't review again the cleaning and refactoring done in bot.js (that is included in this PR too for `loadClock()`)

since all the code is basically black/white copy pasting duplicates, i thought it would be nice to refactor it

did not test it, but it almost not modifies the old behaviour (except 1/2 very minor modifications that shouldnt change the behaviour: for example check am i player for both black and white, instead of black only in the past)

so it should hopefully work as-is or with minor modifications


wonderingabout:
it looks like we are not going to refactor this anytime soon, will open a github issue to keep this in mind if we want to have a look at it again

wonderingabout:
see #276

PR #226, opened 2020-04-11
refactor sai pv

wonderingabout:


wonderingabout:
actually this doesnt replicate the old behaviour faithfully, mistake in constructor passing `setting`

PR #227, opened 2020-04-11
refactor sai score v2

wonderingabout:


wonderingabout:
in better form in #228

PR #213, opened 2019-12-15
functionnal arg checks exports, deprecated ranked / unranked args, refactoring, new features

wonderingabout:
on top of latest devel
(structural changes were merged earlier in #214) 

hopefully will be a lot easier to review now

PR #230, opened 2020-04-12
add r_u_strings, add --challengercolors +/- ranked/unranked, add "all" value to allow all values in allowed families, add increment and range operators in number allowed families

wonderingabout:
taking most of the changes from PR #221 (which itself takes some changes from PR #213 )
Thus, this PR is an intermediate step between devel and PR#221

on top of latest devel

### DIFF
--- a/config.js
+++ b/config.js
@@ -7,7 +7,7 @@ exports.updateFromArgv = function() {
     const ogsPvAIs = ["LeelaZero", "Sai", "KataGo", "PhoenixGo", "Leela"];
 
     const optimist = require("optimist")
-        // 1) ROOT ARGUMENTS:
+        // 1) ROOT ARGUMENTS
         .usage("Usage: $0 --username <bot-username> --apikey <apikey> [arguments] -- botcommand [bot arguments]")
         .demand('username')
         .demand('apikey')
@@ -52,16 +52,16 @@ exports.updateFromArgv = function() {
         .describe('rankedonly', 'Only accept ranked matches')
         .describe('unrankedonly', 'Only accept unranked matches')
         .describe('fakerank', 'Fake bot ranking to calculate automatic handicap stones number in autohandicap (-1) based on rankDifference between fakerank and user ranking, to fix the bypass minhandicap maxhandicap issue if handicap is -automatic')
-        // 2) ARGUMENTS TO CHECK RANKED/UNRANKED CHALLENGES:
-        //     A) ALL/RANKED/UNRANKED FAMILIES:
+        // 2) ARGUMENTS TO CHECK RANKED/UNRANKED CHALLENGES
+        //     2A) ALL/RANKED/UNRANKED FAMILIES
         .describe('bans', 'Comma separated list of usernames or IDs')
         .string('bans')
         .describe('bansranked', 'Comma separated list of usernames or IDs who are banned from ranked games')
         .string('bansranked')
         .describe('bansunranked', 'Comma separated list of usernames or IDs who are banned from unranked games')
         .string('bansunranked')
-        //     B) GENERAL/RANKED/UNRANKED FAMILIES:
-        //         B1) ALLOWED FAMILIES:
+        //     2B) GENERAL/RANKED/UNRANKED FAMILIES
+        //         2B1) ALLOWED FAMILIES
         .describe('boardsizes', 'Board size(s) to accept')
         .string('boardsizes')
         .default('boardsizes', '9,13,19')
@@ -88,7 +88,7 @@ exports.updateFromArgv = function() {
         .default('timecontrols', 'fischer,byoyomi,simple,canadian')
         .describe('timecontrolsranked', 'Time control(s) to accept for ranked games')
         .describe('timecontrolsunranked', 'Time control(s) to accept for unranked games')
-        //         B2) GENERIC GENERAL/RANKED/UNRANKED ARGUMENTS:
+        //         2B2) GENERIC GENERAL/RANKED/UNRANKED ARGUMENTS
         .describe('proonly', 'For all games, only accept those from professionals')
         .describe('proonlyranked', 'For ranked games, only accept those from professionals')
         .describe('proonlyunranked', 'For unranked games, only accept those from professionals')
@@ -215,10 +215,10 @@ exports.updateFromArgv = function() {
     ensureSupportedOgspvAI(argv.ogspv, ogsPvAIs);
 
     // exports arrays
-    // A) general case:
+    // A) general case
     // to be done in PR221
 
-    // B) specific cases:
+    // B) specific cases
     const allowed_r_u_Families_numbers = ["boardsizes",
                                           "komis"
                                          ];
@@ -263,7 +263,7 @@ exports.updateFromArgv = function() {
     }
 
     // r_u exports
-    /* 2) specific r_u cases:*/
+    /* 2) specific r_u cases*/
     for (const familyNameString of ["minrank", "maxrank"]) {
         exportRank(familyNameString, argv);
     }
@@ -276,7 +276,7 @@ exports.updateFromArgv = function() {
     exportBannedFamily(argv);
 
     // console messages
-    // C - test exports warnings:
+    // C - test exports warnings
     testExportsWarnings();
 
     // Show in debug all the ranked/unranked exports results
@@ -293,7 +293,7 @@ function getBLCString(familyNameString, rankedUnranked) {
            + `and/or --${familyNameString}corr${rankedUnranked}`;
 }
 
-// console messages:
+// console messages
 function testDroppedArgv(argv) {
     const droppedArgv = [
          [["botid", "bot", "id"], "username"],
@@ -371,7 +371,7 @@ function ensureSupportedOgspvAI(ogspv, ogsPvAIs) {
     }
 }
 
-// argv.arg(general/ranked/unranked) to exports.(r_u).arg:
+// argv.arg(general/ranked/unranked) to exports.(r_u).arg
 function getArgNameStringsGRU(familyNameString) {
     return ["", "ranked", "unranked"].map( e => `${familyNameString}${e}` );
 }
--- a/config.js
+++ b/config.js
@@ -424,7 +424,8 @@ function exportAllowedFamilyNumbers(familyNameString, argv) {
                 } else if (arg.includes(":")) {
                     const [numberA, numberB, incr] = arg.split(":").map( n => Number(n) );
                     const [min, max] = [Math.min(numberA, numberB), Math.max(numberA, numberB)];
-                    const increment = incr || 1; // default is 1, this also removes arg number 0 (infinite loop)
+                    const increment = Math.abs(incr) || 1; // default is 1, this also removes arg number 0 (infinite loop)
+                                                           // negative increments need to be positive since we force order of min/max
                     // if too much incrementations, sanity check
                     const threshold = 1000;
                     if (((max - min) / increment) > threshold) {
--- a/config.js
+++ b/config.js
@@ -278,14 +278,7 @@ exports.updateFromArgv = function() {
     // console messages
     // C - test exports warnings
     testExportsWarnings();
-
-    // Show in debug all the ranked/unranked exports results
-    if (exports.DEBUG) {
-        const result = JSON.stringify({ ...exports, apikey: "hidden"});
-        console.log(`${"r_u".toUpperCase()} EXPORTS RESULT (apikey hidden):`
-                    + `\n-------------------------------------------------------`
-                    + `\n${result}\n`);
-    }
+    
 }
 
 function getBLCString(familyNameString, rankedUnranked) {
--- a/config.js
+++ b/config.js
@@ -3,38 +3,6 @@
 const fs = require('fs')
 const console = require('console');
 
-exports.check_rejectnew = function() {};
-exports.banned_users = {};
-exports.banned_users_ranked = {};
-exports.banned_users_unranked = {};
-exports.allowed_boardsizes = [];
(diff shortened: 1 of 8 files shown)


config.js
```diff
@@ -240,10 +216,27 @@ exports.updateFromArgv = function() {
                 + `\nDebug status: ${debugStatus}`);
     // B - check unsupported argv //
     testDroppedArgv(argv);
+    checkUnsupportedOgspvAI(argv.ogspv, ogsPvAIs);
```
Dorus: You are no longer calling `toUpperCase` on `argv.ogspv` like you did before. This is okay, as it's a minor detail not needed here, but does mean `checkUnsupportedOgspvAI` has to be modified to include this.

config.js
```diff
@@ -492,10 +416,12 @@ function testDroppedArgv(argv) {
     console.log("\n");
 }
 
-function checkUnsupportedOgspvAI(ogspv, ogsPvAIs) {
-    const upperCaseAIs = ogsPvAIs.map(e => e.toUpperCase());
+function testUnsupportedOgspvAI(ogspv, ogsPvAIs) {
```
Dorus, line 419: possibly call this `validateUnsupportedOgspvAI`?

utils/get_r_u_strings.js
```diff
@@ -0,0 +1,10 @@
+function get_r_u_strings(rankedStatus, speedStatus) {
+    const r_u = rankedStatus ? "ranked" : "unranked";
+    return { r_u,
+             _r_u:              `_${r_u}`,
+             for_r_u_games:     `for ${r_u} games`,
+             for_blc_r_u_games: `for ${speedStatus} ${r_u} games`,
+             from_r_u_games:    `from ${r_u} games` };
+}
+
+exports.get_r_u_strings = get_r_u_strings;
```
Dorus, line 10: File should end in a newline.

utils/get_r_u_strings.js
```diff
@@ -0,0 +1,10 @@
+function get_r_u_strings(rankedStatus, speedStatus) {
+    const r_u = rankedStatus ? "ranked" : "unranked";
```
Dorus, line 2: Can we not just return `r_u` and let the code on the outside add `for` and `from`? That would also eliminate `speedStatus`.

game.js
```diff
@@ -125,13 +126,13 @@ class Game {
                     && !clock.pause.pause_control[`vacation-${clock.black_player_id}`]
                     && !clock.pause.pause_control[`vacation-${clock.white_player_id}`]) {
 
-                    const forRankedUnranked = getForRankedUnranked(this.state.ranked);
-                    const noPauseMg = `Pausing not allowed ${forRankedUnranked}. Resuming game.`;
-                    this.sendChat(noPauseMg);
-                    if (config.DEBUG) this.log(noPauseMg);
+                    const r_u_strings = get_r_u_strings(this.state.ranked);
```
Dorus, line 129: Missing an argument here, but i think the solution would be to remove that argument from that function completely.

utils/get_r_u_strings.js
Dorus: You could then do:
```Js
const noPauseMsg = `Pausing not allowed for ${get_r_u_string(this.state.ranked)} games. Resuming game.`;
```
in other code, instead of the current:
```Js
const r_u_strings = get_r_u_strings(this.state.ranked);
const noPauseMsg = `Pausing not allowed ${r_u_strings.for_r_u_games}. Resuming game.`;
```

Dorus: Oh i see how you use this result in `connection.js`. It makes more sense there and i would keep it, but instead extract the `get_r_u_string` method so you can use just that one in `game.js`.

config.js
wonderingabout: yes :+1:  fixed

config.js
```diff
@@ -492,10 +416,12 @@ function testDroppedArgv(argv) {
...
+    // being case sensitive tolerant
+    const upperCaseOgsPv = ogsPv.toUpperCase();
```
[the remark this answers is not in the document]

config.js
```diff
@@ -506,6 +432,15 @@ function getArgNameStringsGRU(familyNameString) {
     return ["", "ranked", "unranked"].map( e => `${familyNameString}${e}` );
 }
 
+function testNotYetReleasedArgv(argv) {
```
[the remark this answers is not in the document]
wonderingabout: ok, then same as `--publiconly` and `--privateonly`, we'll add these when we need them

config.js
```diff
@@ -272,144 +265,75 @@ exports.updateFromArgv = function() {
...
-    if (argv.boardsizesunranked) {
-        for (const boardsizeunranked of argv.boardsizesunranked.split(',')) {
-            if (boardsizeunranked === "all") {
-                exports.allow_all_boardsizes_unranked = true;
-            } else {
-                exports.allowed_boardsizes_unranked[boardsizeunranked] = true;
+    // r_u exports
+    /* 2) specific r_u cases:*/
+    for (const familyNameString of ["minrank", "maxrank"]) {
+        for (const argNameString of getArgNameStringsGRU(familyNameString)) {
```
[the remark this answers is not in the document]
wonderingabout: we can make it into a function, it removes the nested loop and is simpler, yes

but for the `this.argv`, i dont like it because we dont use the argv except in config, also, we dont pass argv that much so maybe its fine to keep passing argv instead of adding extra things to check about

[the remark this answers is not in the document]
wonderingabout: but then we'd want to do it also with allowed families and other arg exports no ?

config.js
```diff
@@ -272,144 +265,75 @@ exports.updateFromArgv = function() {
...
-    }
-    if (argv.komisunranked) {
-        for (const komiunranked of argv.komisunranked.split(',')) {
-            if (komiunranked === "all") {
-                exports.allow_all_komis_unranked = true;
-            } else if (komiunranked === "automatic") {
-                exports.allowed_komis_unranked[null] = true;
-            } else {
-                exports.allowed_komis_unranked[komiunranked] = true;
+    for (const familyNameString of allowed_r_u_Families_numbers) {
```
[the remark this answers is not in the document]
wonderingabout: in #221 , these `allowed_r_u_Families_numbers` will be at the same position because we use them to filter the args we export (in summary in #221 we dont export any r_u arg (those which have a general arg, a ranked arg, and an unranked arg), the r_u args are exported in a different way after the general arg and ranked/unranked args are filter so that only `config.ranked[familyNameString]` and `config.unranked[familyNameString]` are exported (which are in all cases `config_r_u`))

config.js
```diff
@@ -272,144 +265,75 @@ exports.updateFromArgv = function() {
...
+                        exports[`allow_all_${familyNameString}${_r_u}`] = true;
+                    } else if (familyNameString === "komis" && arg === "automatic") {
+                        exports[`allowed_${familyNameString}${_r_u}`][null] = true;
+                    } else if (arg.includes(":")) {
+                        const [numberA, numberB, incr] = arg.split(":").map( n => Number(n) );
+                        const [min, max] = [Math.min(numberA, numberB), Math.max(numberA, numberB)];
+                        const increment = Math.abs(incr) || 1; // default is 1, this also removes arg number 0 (infinite loop)
+                        // if too much incrementations, sanity check
+                        const threshold = 1000;
+                        if ((Math.abs(max - min) / increment) > threshold) {
```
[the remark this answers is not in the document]
wonderingabout: oh indeed!

config.js
```diff
@@ -272,144 +265,75 @@ exports.updateFromArgv = function() {
...
+                            throw new `please reduce list length in ${argNameString}, max is ${threshold} elements per range.`;
+                        }
+                        for (let i = min; i <= max; i = i + increment) {
```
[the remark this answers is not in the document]
wonderingabout: yes!

config.js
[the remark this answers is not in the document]
Dorus: > but for the this.argv, i dont like it because we dont use the argv except in config, also, we dont pass argv that much so maybe its fine to keep passing argv instead of adding extra things to check about

yes that's why i said a local field for config.js, not one we export.

config.js
```diff
@@ -272,144 +265,75 @@ exports.updateFromArgv = function() {
...
-    if (argv.speedsranked) {
-        for (const e of argv.speedsranked.split(',')) {
-            exports.allowed_speeds_ranked[e] = true;
-        }
-    }
-    if (argv.speedsunranked) {
-        for (const e of argv.speedsunranked.split(',')) {
-            exports.allowed_speeds_unranked[e] = true;
+    for (const familyNameString of (allowed_r_u_Families_strings)) {
+        for (const [argNameString, _r_u] of get_r_u_arr_comma(familyNameString)) {
```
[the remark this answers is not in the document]
wonderingabout: i still wonder if it wouldnt make it too complicated to pass all the args
could we make something like
```
for (const e of arr) {
    exportArg(e, _r_u, argv);
}
```

config.js
wonderingabout: yes its fine

config.js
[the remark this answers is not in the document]
Dorus: > but then we'd want to do it also with allowed families and other arg exports no ?

you mean making it a function we call multiple times? Yes, see my other comments below.

utils/get_r_u_strings.js
wonderingabout: oh i didnt know!
will add :+1:

game.js
Dorus: Scratch that, the solution would be to make a new `get_r_u_string` function with just 1 argument.

utils/get_r_u_strings.js
wonderingabout: yes, its very very convenient in connection.js
but for game.js we use it only once, so we can maybe directly put the result in msg

theres an intermediate way for readability:
```
const for_r_u_games = get_r_u_strings(this.state.ranked);
const noPauseMsg = `Pausing not allowed ${for_r_u_games}. Resuming game.`;
```

game.js
wonderingabout: then i can manually set it to undefined
its really convenient to have one function for all files though, so i'd like to keep it if possible

config.js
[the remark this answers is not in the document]
wonderingabout: i'll try to do it for all functions then :)

config.js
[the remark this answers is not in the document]
Dorus: First step would be to pass many args. Second step would be to figure out how to pass around less args :)

utils/get_r_u_strings.js
Dorus:
```Js
const { for_r_u_games } = get_r_u_strings(this.state.ranked);
```
:D
But i saw you removed the need for this function entirely from the pause code. Nice!

wonderingabout: yeees i'm so smart thanks to your teachings and your guidance xD

config.js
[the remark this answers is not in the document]
wonderingabout: i was considering the loop variant to pass args only once loool

config.js
```diff
@@ -216,7 +212,7 @@ exports.updateFromArgv = function() {
                 + `\nDebug status: ${debugStatus}`);
     // B - test unsupported argv //
     testDroppedArgv(argv);
-    testUnsupportedOgspvAI(argv.ogspv, ogsPvAIs);
+    validateUnsupportedOgspvAI(argv.ogspv, ogsPvAIs);
```
Dorus, line 215: Actually `validateOgspvAI` would cover the functionality just fine too.

config.js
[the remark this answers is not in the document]
wonderingabout: i fixed it like this:
https://github.com/online-go/gtp2ogs/pull/230/commits/0fe038af02b2297d23b070ef580319d585a47733
in pr #221 it will make sense to keep the loops here because we'll filter args r_u and not export them automatically

config.js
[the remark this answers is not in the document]
wonderingabout: same as https://github.com/online-go/gtp2ogs/pull/230#discussion_r415144444

config.js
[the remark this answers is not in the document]
wonderingabout: same as https://github.com/online-go/gtp2ogs/pull/230#discussion_r415144444

docs/OPTIONS-LIST.md
```diff
@@ -284,9 +285,11 @@ For the allowed families arguments, you can either use the value:
 comma-separated values (without space) will allow every value inputted,
 every other value will be rejected
 - for numbers +/- text families (5.5,6.5,7.5 (komis), 9,13,19
-(boardsizes)), it is possible to use as well the "range"
+ (boardsizes)), it is possible to use as well the "range"
 operator `:` to navigate one by one from min to max (ex:
-`5.5:7.5` is `5.5,6.5,7.5` and `13:17` is `13,14,15,16,17`).
+ `5.5:7.5` is `5.5,6.5,7.5` and `13:17` is `13,14,15,16,17`),
+ as well as the "increment" operator (ex: `13:19:2` is `13,15,17,19`,
+see [notes A-](/docs/NOTES.md#a) for details.
```
Dorus, line 292: Do we need this dash after the A?

wonderingabout: no indeed, there was a dash because the linked used to be in the `/docs/NOTES.md/a-`, because the category title was `## A:`, but now since we removed the `:` we dont need the `a-` anymore, so we can just remove as well the `A-` 's dash

wonderingabout:
this is too big and all over the place, not going to be merged,
there are many good parts to take from it, will be shrinked into smaller prs on top of latest devel

wonderingabout:
see #279

PR #231, opened 2020-04-14
Minmax common function

wonderingabout:
on top of #230


wonderingabout:
this needs to be made into a specific pr that modfies only the minmax code in connection.js, not on top of other prs
consider shrinking it in smaller chunks if too long

PR #232, opened 2020-04-16
add --whitelist and --blacklist, deprecate --bans

wonderingabout:
on top of #221


wonderingabout:
this is untested and includes a lot of other uneeded changes
need to make a new one with only the whitelist change on top of latest devel

wonderingabout:
see #280

PR #233, opened 2020-04-19
deprecate boardsizeswidths/heights, update documentation

wonderingabout:
this is a much lighter version of #230 , focusing mostly on documentation and deprecation changes


docs/NOTES.md
```diff
@@ -110,9 +93,13 @@ always returns `-1` regardless of actual handicap stone number
...
-- `--fakerank 6d` and `--maxhandicap 4` and user ranking 
-`2k`
-- 
+`--fakerank 6d --maxhandicap 4` and user ranking `2k`:
+
+- Automatic handicap stones: 6d - 2k = 8 rank difference 
+=> 8 automatic handicap stones
+- but max handicap is 4 stones
+- 8 (detected automatic stones wanted by user) > 
+max stones allowed 4 => challenge rejected
```
Dorus, line 102: For line 91-102:
```
Currently, when handicap is automatic, ogs does not inform us what the handicap
is before accepting the challange. Also, challange notifications do not contain
our rank. With fakerank we can estimate the handicap based on the oponent
ranking and our (fake) rank.

Example use case:
`--fakerank 6d --maxhandicap 4` and user ranking `2k`:
Expected handicap is 6d-2k = 8 ranks. 8 ranks > 4 max handicap => Challange rejected.
```

docs/OPTIONS-LIST.md
```diff
@@ -199,6 +199,10 @@ Below are the "only" options that only have a general option
 
   ```--unrankedonly```  Only accept unranked matches
 
+  ```--privateonly```  (**Not yet supported**) Only accept private matches
+
+  ```--publiconly```  (**Not yet supported**) Only accept public (non-private) matches
```
Dorus, line 204: If this is not supported, we should not add it.

config.js
```diff
@@ -458,64 +422,71 @@ exports.updateFromArgv = function() {
...
+         [["speedranked"], "speedsranked"],
+         [["speedunranked"], "speedsunranked"],
+         [["timecontrol"], "timecontrols"],
+         [["timecontrolranked"], "timecontrolsranked"],
+         [["timecontrolunranked"], "timecontrolsunranked"]
+    ];
+    for (const [oldNames, newName] of deprecatedArgv) {
+        for (const oldName of oldNames) {
+            if (argv[oldName]) {
+                console.log(`Deprecated: --${oldName} is no longer `
```
Dorus, line 479: `Deprecated` usually means that a feature will be removed in future released, we should probably not use that word unless we go around and fix the config to use the new value (until we actually drop support sometime in the future?)

Anyway I'm ok with not supporting these arguments, but if we do we should not use the word "deprecated", instead stick to just "unsupported/no longer supported" etc.

wonderingabout: aaaah important nuance!!!!

so `deprecated` refers to code that is still existing and working but not supported anymore!

docs/NOTES.md
wonderingabout: its clearer indeed, i'll add it

docs/OPTIONS-LIST.md
wonderingabout: @anoek said that he may take some time in the future to add it in notification, and several bot admins said that they wanted to use this feature

if this feature is added soon in the future, it will be easier to toggle it on

config.js
Dorus: Actually it means: Still supported right now, but support will be dropped in the future.
That does usually mean no "support" support, like no fixes to this broken feature. But also means current code that uses this should keep working.

PR #236, opened 2020-04-19
Markup improvements.

Dorus:
Need to double check if i didn't modify something my mistake (i had some merge whoes). Should be layout and markup mostly, and some text moved around.


Dorus:
@roy7  Meta said we can merge this.

PR #238, opened 2020-04-20
(unfinished rewrite) add --opentimesweek and --opentimecountdown +/- ranked/unranked

wonderingabout:
so that we can leave with peace of mind our bots auto handle shutdown time for everyday of the week :100:


wonderingabout:
update: i think i'll rewrite this code in a way that it calculates countdown until shutdown (or before opening)

this is because, for example if we want to close our bot at 1pm then this code wont allow it to open the bot for all the day after 1pm

wonderingabout:
update: i think i'll rewrite this code by having periods of being open time rather than min and max

this is because, for example if we want to close our bot at 1pm then this code wont allow it to open the bot for all the day after 1pm

by specifiying periods of open time everyday, it is possible to repoen the bot after closing it

wonderingabout:
see #238

PR #239, opened 2020-04-25
(untested) move some shared functions to /utils/

wonderingabout:
did not test it yet

### DIFF
--- a/utils/decodeMoves.js
+++ b/utils/decodeMoves.js
@@ -1,0 +1,101 @@
+const { char2num } = require("./utils/char2num");
+
+function pretty_char2num(ch) {
+    if (ch === ".") return -1;
+    return "abcdefghjklmnopqrstuvwxyz".indexOf(ch.toLowerCase());
+}
+
+function decodeMoves(move_obj, board_size) {
+    const ret = [];
+    const width = board_size;
+    const height = board_size;
+
+    /*
+    if (DEBUG) {
+        console.log("Decoding ", move_obj);
+    }
+    */
+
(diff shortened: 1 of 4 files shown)


utils/decodeMoves.js
```diff
@@ -0,0 +1,101 @@
+const { char2num } = require("./utils/char2num");
```
Dorus, line 1: You are already in the `./utils` folder. Correct import would be `./char2num`

Dorus: I really wonder if you ever try to run code before you open a pr meta, travis spotted this instantly.

wonderingabout: indeed, its been around 2 months (since time i moved to ubuntu 20.04, that didnt have npm support until a few days ago) that i stopped running gtp2ogs at all
but it will change soon now, sorry about that :+1:

wonderingabout: also i didnt know exports worked like that, now i see how it works :)

Dorus: Yes, imports work with relative path.

There are ways, at least in typescript, to also set some root folders so you can always refer them the same way. It's possible with some Js sompiles too i see, so you would get things like `~/utils/char2num` or, in TS when you set a path "utils" `@utils/char2num`.

wonderingabout: yes, i thought we were always referring to absolute path, but now i see :)

wonderingabout:
this is now out of sync, should make a new pr

PR #241, opened 2020-04-26
--aichat option allowing for posting chat to discussion/malkovich from stderr

sanderland:
All fixed, with a new option.
Also fixed a bug where not having --ogspv would cause an error.


config.js
```diff
@@ -36,7 +36,7 @@ exports.allowed_timecontrols_ranked = {};
 exports.allowed_timecontrols_unranked = {};
 
 exports.updateFromArgv = function() {
-    const ogsPvAIs = ["LeelaZero", "Sai", "KataGo", "PhoenixGo", "Leela"];
+    const ogsPvAIs = ["LeelaZero", "Sai", "KataGo", "PhoenixGo", "Leela","Free"];
```
roy7, line 39: Could you add a space after the comma just for formatting to be consistent? :)

pv.js
```diff
@@ -63,7 +66,13 @@ class Pv {
             this.pvLine = null;
         }
     }
-
+    postPvToChatFree(errline) {
+        const stop = this.STOPRE.exec(errline);
+        if (stop && this.pvLine) {
```
Dorus, line 71: How does this work? `this.pvLine` is always false for me.

sanderland: it doesn't work, I made a draft PR to discuss

pv.js
```diff
@@ -63,7 +66,13 @@ class Pv {
...
+            const body = this.getPvChat(stop);
+            this.game.sendChat(body, "malkovich");
```
Dorus, line 73: The `sendChat` function has 3 parameters, body, move number and type (malkovich, or the default, discussion). You are sending the string malkovich as move number now, ogs does not understand this.

Either send no move number, or be nice and send the current move as move number. Either way make "malkovich" the third parameter if you intend to send this to malkoch log. if not you could just call `this.game.sendChat(body, this.game.state.moves.length + 1)` instead.

The other functions have  +1 on the moves to make the pv land on the correct move. We probably want that here too.

pv.js
Dorus: If you remove the check for `&& this.pvLine`, or if you set `this.pvLine` to a non-falsy value, like done in `postPvToChatSingleLine`, it will work.

Dorus:
I've checked the code, and it looks mostly ok. I do think you tried a little too hard to replicate the existing pattern.

For example the `getPvChatFree` function is a bit simple. Since `postPvToChatFree` is used only for free pv you can inline it there. The only reason we split this up before, is because we used  `postPvToChatSingleLine` and `postPvToChatDualLine` for multiple bots.

Another option would be to reuse `postPvToChatSingleLine` instead, that would include all functionality including the ignore pondering output and compatibility with `postPvToChatLastLine`.

sanderland:
Thanks for the pointers, will try to fix/clean later. Copy-paste jank tends to happen when you don't know the codebase nor the language very well. ;)

roy7:
Seems simple and straight forward to me now. If we get @Dorus blessing we can merge it. :)

wonderingabout:
i dont know about what this code technically does, but for documentation

you'd also need to add the option in OPTIONS-LIST.md :)
https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md
(copy paste description from config.js argv `aibot`)
(and just update the existing ogspv notes to add `Free` as well as a quick explanation so that even new users can understand what it does)

and possibly NOTES.md too
https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md
and also add Free in the documentation of ogspv

wonderingabout:
also we dont want to use `var`, prefer `let` or even better `const` if possible
https://github.com/online-go/gtp2ogs/pull/241/files#diff-9d199097bf1be7e582b7cb3a378cfbf0R50

bot.js
```diff
@@ -38,15 +39,19 @@ class Bot {
...
-
         this.proc.stderr.pipe(split2()).on('data', (data) => {
             if (this.ignore)  return;
             const errline = data.toString().trim();
             if (errline === "") return;
             this.error(`stderr: ${errline}`);
 
             if (config.ogspv) this.pv.postPvToChat(errline);
+            if (config.aichat) {
+               var chat_match = /(DISCUSSION|MALKOVICH):(.*)/.exec(errline)
```
wonderingabout, line 50: also we dont want to use `var`, prefer `let` or even better `const` if possible
https://github.com/online-go/gtp2ogs/pull/241/files#diff-9d199097bf1be7e582b7cb3a378cfbf0R50

sanderland: fixed

config.js
```diff
@@ -487,12 +488,14 @@ function testDroppedArgv(argv) {
 
 function ensureSupportedOgspvAI(ogspv, ogsPvAIs) {
     // being case tolerant
-    const upperCaseOgsPv = ogspv.toUpperCase();
-    const upperCaseAIs   = ogsPvAIs.map(e => e.toUpperCase());
+    if(ogspv) {
```
wonderingabout, line 491: it is a good idea to add the `if`, but i think we would want to make the `if` outside of the function, and run the function only if the arg is used, what do you think @Dorus 

```Js
if (argv.ogspv) ensureSupportedOgspvAI(ogspv, ogsPvAIs);
```

wonderingabout:
hope i'm not being too picky :)

config.js
sanderland: fixed

sanderland: it's not just a good idea, it crashed the damn program when I used it :o

config.js
```diff
@@ -240,7 +241,9 @@ exports.updateFromArgv = function() {
                 + `\nDebug status: ${debugStatus}`);
     // B - test unsupported argv //
     testDroppedArgv(argv);
-    ensureSupportedOgspvAI(argv.ogspv, ogsPvAIs);
+    if (argv.ogspv) {
```
wonderingabout, line 244: i'd consider removing nesting and just 

Js
```
if (argv.ogspv) ensureSupportedOgspvAI(argv.ogspv, ogsPvAIs);
```

docs/OPTIONS-LIST.md
```diff
@@ -133,6 +133,17 @@ for example `--ogspv LeelaZero`
 
 **IMPORTANT: see [notes H](/docs/NOTES.md#h) for details**
 
+### aichat
+
+`--aichat` Allow bots to send messages to the chat and malkovich log.
+
+Bots should output messages to standard error output in the format:
+
+```
```
wonderingabout, line 142: you need to add the flag `Text`, mardownlint enforces to always have a language flag for each code quotations blocks

config.js
Dorus, line 244: Why is this required? Does this if fix bug where `ensureSupportedOgspvAI` cannot handel `argv.ogspv==null`? If so, i would prefer to fix it inside the `ensureSupportedOgspvAI` function.

bot.js
```diff
@@ -38,15 +39,19 @@ class Bot {
...
-
         this.proc.stderr.pipe(split2()).on('data', (data) => {
             if (this.ignore)  return;
             const errline = data.toString().trim();
             if (errline === "") return;
             this.error(`stderr: ${errline}`);
 
             if (config.ogspv) this.pv.postPvToChat(errline);
+            if (config.aichat) {
+                const chat_match = /(DISCUSSION|MALKOVICH):(.*)/.exec(errline)
```
Dorus, line 50: What bots support this function? Also why does it not hook into pv code? `pv` object allows `postPvToChat` to be replaced by a function that could just does what line 50-53 does here.

The code itself does look sound so if you think there is value in supporting both pv and this discussion option then it absolutely has to be done like this. I'm fine either way.

Dorus: To explain the pv code: It should have been an interface that has just `postPvToChat` where you hook up the correct implementation based on the requested bot. However because my limited Js knowledge on how to make interfaces in Js (with Ts i would totally have done that haha), we ended up with 1 pv class that modifies itself based on the requested bot.

config.js
sanderland: const upperCaseOgsPv = ogspv.toUpperCase();
undefined.toUpperCase is not defined
I had fixed it in there, and was then told to move it here.

bot.js
sanderland: It's interesting that you think gtp2ogs should be made to support certain bots, rather than someone making a bot to use with gtp2ogs. I've done the latter, and found this handy. Other bots may use it as well.
I am not sending PVs, and PVs are broken at OGS when saving SGFs.

Dorus: yeah i guess that makes sens. All ok then.

config.js
```diff
@@ -487,12 +488,14 @@ function testDroppedArgv(argv) {
 
 function ensureSupportedOgspvAI(ogspv, ogsPvAIs) {
     // being case tolerant
-    const upperCaseOgsPv = ogspv.toUpperCase();
-    const upperCaseAIs   = ogsPvAIs.map(e => e.toUpperCase());
+    if (ogspv) {
```
Dorus, line 491: I would reverse this:
```Js
if (!ogspv) return;
```
that way you avoid deep nesting.

config.js
```diff
@@ -488,14 +488,13 @@ function testDroppedArgv(argv) {
 
 function ensureSupportedOgspvAI(ogspv, ogsPvAIs) {
     // being case tolerant
-    if (ogspv) {
-        const upperCaseOgsPv = ogspv.toUpperCase();
-        const upperCaseAIs   = ogsPvAIs.map(e => e.toUpperCase());
+    if (!ogspv) return
```
Dorus, line 491: I'm missing a semicolon here.

Dorus:
@roy7 I'm ok with a merge here.

PR #242, opened 2020-04-26
Add markdown rules to lint.

Dorus:
Added markdown rules to project so travis can run them too.


wonderingabout:
very very nice!!!!!!

PR #243, opened 2020-04-27
add DEV.md, quick fixes and minor updates

wonderingabout:


connection.js
```diff
@@ -691,8 +691,8 @@ function getArgNameStringsGRU(familyNameString) {
 
 function rankToString(r) {
     const R = Math.floor(r);
-    if (R >= 30)  return `${R - 30 + 1}d`; // R >= 30: 1 dan or stronger
```
Dorus, line 694: This was clearer with the `:` in i think.

Dorus:
This one is ok to merge after my 1 comment.

connection.js
wonderingabout: ok, i wanted to remove as much punctuation as possible like markdown lint does, but it is indeed clearer in that case :1st_place_medal:

wonderingabout: fixed

PR #245, opened 2020-04-27
(untested) move check_rejectnew duplicate export to test.js

wonderingabout:


Dorus:
Why do we need this?

wonderingabout:
oops, this PR intended to also remove `check_rejectnew export` from config.js here:
https://github.com/online-go/gtp2ogs/blob/devel/config.js#L6

because we already export it in https://github.com/online-go/gtp2ogs/blob/devel/config.js#L269-L274

the error if we try to remove line 6 from config.js is not due to gtp2ogs behaviour, its a specific issue of test.js which doesnt fetch args from config.js, thats why its more accurate to remove line 6 from config.js (uneeded) and put it in test.js (where it is needed), same as what was done for https://github.com/online-go/gtp2ogs/blob/devel/test/test.js#L28-L29

wonderingabout:
better addressed in #289

PR #246, opened 2020-04-27
(untested) export allowed families in function body

wonderingabout:


wonderingabout:
better addressed in #297

PR #248, opened 2020-04-27
function renaming with verbs

wonderingabout:


wonderingabout:
we should rename functions names with a verb (that describes the action done) + the rest of the function title that describes what is done in a self explanatory way

parameters renaming should be done in a separate pr, as it obviously creates merge conflicts with other changes too

and finally, there are many parts of the code that we could rework

PR #249, opened 2020-04-27
(untested) update getBannedFamilyReject

wonderingabout:
this matches #230 
did not test it yet


wonderingabout:
this is not compatible with #204 , closing it and maybe we'll submit an enhanced version in the future

PR #250, opened 2020-04-27
(untested) update check booleans r_u

wonderingabout:


wonderingabout:
this is out of sync and was done in a better way in #287

PR #251, opened 2020-04-27
(untested) tweak non-square boardsize reject

wonderingabout:


wonderingabout:
this is too out of sync with devel and #287 , should be redone directly there if still relevant

PR #252, opened 2020-04-27
(untested) use objects for allowed families numbers, refactor checkChallengeAllowedFamilies

wonderingabout:


wonderingabout:
this is too out of sync and should be kept in mind in #305 

especially if we want to support non square boardsizes, using a "matrix-like" object check (ex; 25x1 `allowed_boardsizes[25][1] || allowed_boardsizes[1][25]`

PR #254, opened 2020-05-01
Notify bot of end of the game using final_score and send it to chat

sanderland:
ignore the commit trail, didn't checkout properly, diff should be small.

This was motivated by me wanting to save SGFs of all games my bots play, so I looked for a sensible gtp command.


config.js
```diff
@@ -49,6 +49,7 @@ exports.updateFromArgv = function() {
         .string('greeting')
         .describe('farewell', 'Thank you message to appear in chat at end of game (ex: -Thank you for playing-)')
         .string('farewell')
+        .describe('farewell_score', 'Send the score according to the bot at the end of the game')
```
wonderingabout, line 52: so far no gtp2ogs option uses an underscore `_`, we may want to keep it as that, but @Dorus and @roy7 most likely know better than me

game.js
```diff
@@ -454,11 +453,24 @@ class Game {
...
+                if (this.bot) { // only kill the bot after it processed this
+                    this.bot.gameOver();
+                    this.ensureBotKilled();
+                }
+            };
+            this.bot.command('final_score', sendTheScore, false, true); // allow bot to process end of game
+        }
+        else {
+            if (this.bot) {
+                this.bot.gameOver();
```
wonderingabout, line 470: this code is duplicated

```Js
                if (this.bot) { // only kill the bot after it processed this
                    this.bot.gameOver();
                    this.ensureBotKilled();
                }
```

you may want to add a method (or a function instead) and do 2 calls

wonderingabout: also the nested `else` then `if` could be written in one `else if` i think, even though it may not be best way to code this

game.js
```diff
@@ -354,7 +354,6 @@ class Game {
             'game_id': this.game_id,
             'move': encodeMove(move)
         }));
-        //this.sendChat(`Test chat message, my move #${move_number}  is: ${move.text}`, move_number, "malkovich");
```
wonderingabout, line 357: why do we remove this line?

config.js
wonderingabout: also, if the purpose is just to send the score, your bot should already be set up to sent that information at the last move of `config.aichat`

if the purpose is to send a specific bot chat at the end of the game, you may want to use a more general `farewellaichat` option to send any text, not just score
in game.js you'd just have to send this text before killing bot

game.js
sanderland: It seemed like a leftover test that should be.

config.js
sanderland: The bot doesn't know it's the last move, since it gets murdered on a gameOver (resign) event.
and quit is ambiguous as well.

game.js
Dorus: yeah small improvements like this are ok. They do slow down the review process, so separate them if it becomes too big. (but that holds true for any pr)

game.js
```diff
@@ -454,11 +453,24 @@ class Game {
         const winloss = (this.state.winner === this.conn.bot_id ? "W" : "L");
         this.log(`Game over.   Result: ${col}+${res}  ${winloss}`);
 
-        if (this.bot) {
-            this.bot.gameOver();
-            this.ensureBotKilled();
+        // Notify bot of end of game and send score
+        if (config.farewell_score && this.state) {
```
Dorus, line 457: Why do we need this check on `this.state`? Also we've done `this.state.winner` and a few more property reads a few lines before, so i don't think we can reach this line with `this.state` is `null` or `false`.

sanderland: this is basically  because farewell does, and i am not sure what leads to state being null

sanderland: fixed

game.js
sanderland: fixed

game.js
```diff
@@ -454,7 +453,18 @@ class Game {
         const winloss = (this.state.winner === this.conn.bot_id ? "W" : "L");
         this.log(`Game over.   Result: ${col}+${res}  ${winloss}`);
 
-        if (this.bot) {
+        // Notify bot of end of game and send score
+        if (config.farewellscore && this.bot) {
+            const sendTheScore = (score) => {
+                if (score) this.log(`Bot thinks the score was ${score}`);
+                if (res !== "R" && res !== "Time" && res !== "Can") this.sendChat(`Final score was ${score} according to the bot.`, "discussion");
```
Dorus, line 460: Why this check? It wasn't required in the last line, and also why not check for score like in the previous line?

Shouldn't the check for send chat and log be equal?

Dorus: Also i do not think res can change here, so we could check for res before we send the bot the score command and skip it all together in that case.

Dorus: Also i do not think res can change here, so we could check for res before we send the bot the score command and skip it all together in that case.

Dorus: Also i do not think res can change here, so we could check for res before we send the bot the score command and skip it all together in that case.

Dorus: Also i do not think res can change here, so we could check for res before we send the bot the score command and skip it all together in that case.

sanderland: resign, timeout or game cancelled.
as I said, my major motivation for this was a signal that the game has ended, the chat is kind of an aside.

Dorus: So if you do not send the chat message for resign, cancel or timeout, what remains where you do send it?

sanderland: double pass, where score actually makes sense. but in all cases the bot knows the game has ended, and can save it's own log, which is frankly the main thing I am using this for. Happy to rename it, and/or remove the chat.

game.js
```diff
@@ -454,7 +453,18 @@ class Game {
...
+                if (this.bot) { // only kill the bot after it processed this
+                    this.bot.gameOver();
+                    this.ensureBotKilled();
+                }
+            };
+            this.bot.command('final_score', sendTheScore, false, true); // allow bot to process end of game
```
Dorus, line 466: use `null` for missing `eb`, not `false`.

game.js
Dorus: 1. I think it's interesting to show final score on resign too. In fact on double pass, ogs score is usually accurate.
2. is it possible to detect and whitelist double pass instead of excluding time and cancel? Seems safer to use a whitelist for the things we want to show, instead of blacklist some. That way we also wont act on unknown values in case ogs adds something new or we forgot an option.

sanderland: 1. done . 2. not easy and I think it's not too bad if we send on some unknown value.

PR #255, opened 2020-05-01
add automatic handicap -1 warning

wonderingabout:


wonderingabout:
should be in an issue first before we decide if we want to add this

wonderingabout:
see #28

PR #264, opened 2020-05-10
Count games even before it's our move.

Dorus:


connection.js
```diff
@@ -505,10 +500,10 @@ class Connection {
...
     }
-    processMove(gamedata) {
-        const game = this.connectToGame(gamedata.id)
-        game.makeMove(gamedata.move_number);
-    }
+    // processMove(gamedata) {
+    //     const game = this.connectToGame(gamedata.id)
+    //     game.makeMove(gamedata.move_number);
+    // }
     processStoneRemoval(gamedata) {
```
wonderingabout, line 507: unless i'm mistaken you can also delete or comment out processStoneRemoval

wonderingabout: on_friendRequest also seems unused:
https://github.com/Dorus/gtp2ogs/blob/CountActiveGames/connection.js#L260-L266

Dorus: Indeed, so many unused functions here. Same with game started method. And delete too if i add it to the ignore list on top.

I think i'm going to make another pr for delete notifications and add this there.

wonderingabout: yes, making a separate pr would be better, here we're just fixing the connectedgames problem

wonderingabout: you may want to put back these lines as they were, as i made a specific pr #266

roy7:
@Dorus This good to merge?

Issue #256, opened 2020-05-02
Extra gamedata package leading to incorrect genmove


sanderland:
I'm seeing my bot getting a genmove command before all the play commands are finished. possibly there is a race condition somewhere.
```
May 02 21:12:32   game.js:496                    [Game 23604177]  Move 2  Got     d4
May 02 21:12:32   bot.js:148                     [21705] Calling sendMove with d4
May 02 21:12:32   bot.js:148                     [21705] >>> play white d4
May 02 21:12:32   game.js:496                    [Game 23604177]  Move 2  makeMove 2 is 2 !== 2 ?
May 02 21:12:32   bot.js:148                     [21705] Generating move for game 23604177
May 02 21:12:32   game.js:496                    [Game 23604177]  Move 2  genmove black
May 02 21:12:32   bot.js:148                     [21705] >>> genmove black
May 02 21:12:32 !!!!! bot.js:156                     [21705] stderr: GOT INPUT genmove black
May 02 21:12:32 !!!!! bot.js:156                     [21705] stderr: P:Tenuki generating move
May 02 21:12:32 !!!!! bot.js:156                     [21705] stderr: Generated move Move(WD4)
May 02 21:12:32 !!!!! bot.js:156                     [21705] stderr: dscore -0.0838177959694395 = 0.8592221672877765 0.943039963257216 at Move(WD4)...
May 02 21:12:32 !!!!! bot.js:156                     [21705] stderr: GOT INPUT play white d4
```

roy7:
Could you provide all of the log?

sanderland:
I don't have the full log, usually this ends up in a sea of text and it's rare I catch it like this

roy7:
To summarize some thoughts from our Discord conversation:

The writes are done via the command() function which should be issuing to the bot in order because it just does a simple this.proc.stdin.write().

sendMove() simply calls this.command().

Is it possible your bot is reading stdin multi-threaded and there's a race condition on the bot side?

There's nothing async in command() or sendMove() so in theory the gtp2ogs log is correct and the 'genmove' is going out stdin.write() after the 'play'.

Is this linux or Windows? We do have an issue with Windows running node that Leela seemingly can't use pondering unless we add a ~200ms delay between each command, I assumed somehow the pipe was losing info or leela was failing to read stdin, but maybe commands get mixed out of order and that is what breaks leela pondering. If you are on linux that isn't a concern.

sanderland:
[...]

PR #265, opened 2020-05-10
Move gamedataChanged test before this.state is overwritten

roy7:
Noticed this when trying to debug https://github.com/online-go/gtp2ogs/issues/256

Was this way since original PR https://github.com/online-go/gtp2ogs/pull/193

### DIFF
--- a/game.js
+++ b/game.js
@@ -41,7 +41,10 @@ class Game {
 
             //this.log("Gamedata:", JSON.stringify(gamedata, null, 4));
 
+            // Test for chanaged state before overwriting prior state with gamedata packet
+            const gamedataChanged = (this.state ? JSON.stringify(this.state) !== JSON.stringify(gamedata) : true);
             const prev_phase = (this.state ? this.state.phase : null);
+
             this.state = gamedata;
             this.my_color = this.conn.bot_id === this.state.players.black.id ? "black" : "white";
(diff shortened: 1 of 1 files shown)


game.js
```diff
@@ -41,7 +41,10 @@ class Game {
 
             //this.log("Gamedata:", JSON.stringify(gamedata, null, 4));
 
+            // Test for chanaged state before overwriting prior state with gamedata packet
```
wonderingabout, line 44: small typo there `changed` :sweat_smile:

roy7: Thanks.

sanderland:
could this be a clearer check at the top like
```
if this.state
  if equal, return / ignore
  if not, restart bot
```
or is doing half the function somehow necessary?

roy7:
It could be changed that way. This is just how it was originally added so noone ever changed it. What your change would so is mean we don't need to check if the bot is running if we already knew this.state was null (since a null state can't have a running bot, right?). 

We could perhaps move it around now, or in a separate PR. Since this PR needs updated tests, I'm leaning towards minimal changes here to update the functionality and we can reformat later?

roy7:
I'm referring to https://github.com/online-go/gtp2ogs/pull/268 which make this PR obsolete.

PR #267, opened 2020-05-10
Deletenotifications

Dorus:
On top of #264 

Delete notifications so that the bot interface stays clean. Else all these notifications are transmitted to the bot each time it connect, and that can be many.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -14,6 +14,7 @@ const Game = require('./game').Game;
 /** Connection **/
 /****************/
 const ignorable_notifications = {
+    'delete': true,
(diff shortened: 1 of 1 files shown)


wonderingabout:
nice refactoring, i really like it

roy7:
@Dorus I think it's just some code that was removed elsewhere is still present here, if you want to update to latest devel?

Dorus:
This was on top of #264 and when we pr that one, it longer recognized the right commit order.

I rebased and now it's fixed.

PR #268, opened 2020-05-10
If we get a duplicate gamedata and nothing has changed, ignore it

roy7:
This can happen if connection timeout is slow and we re-request gamedata, but it was already on the way?

Timeouts were added in https://github.com/online-go/gtp2ogs/commit/9693202d2dc91ceeeaf812429ddddcad7473c257 but it seems the errors in https://github.com/online-go/gtp2ogs/issues/256 were when we re-requested gamedata and got it a second time, but issued a move command to server before processing the 2nd gamedata packet.

### DIFF
--- a/game.js
+++ b/game.js
@@ -39,6 +39,16 @@ class Game {
         this.socket.on(`game/${game_id}/gamedata`, (gamedata) => {
             if (!this.connected) return;
 
+            const gamedataChanged = false;
+
+            if (this.state) {
+                gamedataChanged = (JSON.stringify(this.state) !== JSON.stringify(gamedata));
+
+                // If the gamedata is idential to current state, it's a duplicate. Ignore it and do nothing.
+                this.log('Ignoring gamedata that matches current state');
+                if (!gamedataChanged) return;
+            }
+
             //this.log("Gamedata:", JSON.stringify(gamedata, null, 4));
 
             const prev_phase = (this.state ? this.state.phase : null);
@@ -80,8 +90,6 @@ class Game {
             // restart the bot by killing it here if another gamedata comes in. There normally should only be one
             // before we process any moves, and makeMove() is where a new Bot is created.
             //
-            const gamedataChanged = (JSON.stringify(this.state) !== JSON.stringify(gamedata));
-
             if (this.bot && gamedataChanged) {
                 this.log("Killing bot because of gamedata change after bot was started");
 
--- a/game.js
+++ b/game.js
@@ -41,7 +41,7 @@ class Game {
 
             const gamedataChanged = this.state ? (JSON.stringify(this.state) !== JSON.stringify(gamedata)) : false;
 
-            if (this.state & !gamedataChanged) {
+            if (this.state && !gamedataChanged) {
                 // If the gamedata is idential to current state, it's a duplicate. Ignore it and do nothing.
                 this.log('Ignoring gamedata that matches current state');
                 return;
--- a/game.js
+++ b/game.js
@@ -41,8 +41,9 @@ class Game {
 
             const gamedataChanged = this.state ? (JSON.stringify(this.state) !== JSON.stringify(gamedata)) : false;
 
-            if (this.state && !gamedataChanged) {
-                // If the gamedata is idential to current state, it's a duplicate. Ignore it and do nothing.
+            if (this.state && !gamedataChanged && this.bot && !this.bot.dead) {
+                // If the gamedata is idential to current state, it's a duplicate. Ignore it and do nothing, unless no
+                // bot is running.
                 this.log('Ignoring gamedata that matches current state');
                 return;
             }
--- a/game.js
+++ b/game.js
@@ -39,14 +39,12 @@ class Game {
         this.socket.on(`game/${game_id}/gamedata`, (gamedata) => {
             if (!this.connected) return;
 
-            let gamedataChanged = false;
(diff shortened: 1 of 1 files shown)


roy7:
Makes https://github.com/online-go/gtp2ogs/pull/265 obsolete.

roy7:
@sanderland gtp2ogs needs correct state set up before running the makeMove. The question is if the bot needs restarted (import fresh copy of this.state) or not. I'm not sure the objection. Do you want to kill the bot earlier before we do the state setup instead of where it is now?

Dorus:
How do we not notice that we got the same move twice? we should be able to request the same move 10x and still handle it properly right? I would rather not have us depend on (lack of) server lag, and blindly increasing the timeout could eat up a lot of time for the bot in live games.

roy7:
It isn't about the move, it's the gamedata. When the timeout fires after 1 second and do a socket level disconnect/connect to force a new gamedata to be sent, that might happen after we already sent a Move command to the server. But the gamedata is coming back already. So when we process it, it doesn't have the move we sent to the server. And we're out of sync on our new genmove. If we don't restart the bot (like we were) we might try moving twice in a row or moving to a different spot and having no idea our board doesn't match server's board.

In proper operation gamedata should only be sent 1 time ever. It might happen a 2nd time during a server restart, I think that's why I originally added the 'restart bot if we get unexpected gamedata'. Meaning another gamedata after we already had state. @windo though sometimes had problems where he never got a gamedata packet in the first place. Perhaps also during server restart? And thus his bot would hand forever and time out. So that's why he added his timeout code to force a gamedata re-send.

I'm just thinking 1 second on that is too fast, we're really only trying to capture a true server error situation (gamedata was never sent to us so we have no state even though we connected to the game). If someone is across the world and it takes over 1 second for gamedata to arrive, we've not triggered a timeout to get it re-sent and that might be while a move is already processing or was recently sent to the server (after we sent the gamedata request, but before the move was processed by server, so the gamedata we get back is wrong, a move command will follow after it, but it's too late since we thought it was our turn and we're already called genmove).

roy7:
I'm guessing why others never had this happen but @sanderland is, is maybe his server's connection to OGS is more laggy so he's generating these extra gamedata packets and we aren't. If you want to try and reproduce it, you could set timeout to .1 second and use test server and see what happens.

roy7:
I'm unsure if the test needs updated, or if this really is an error state and the test is saving the day.

I wonder if a connect event should clear any state for that game. The restarts might be failing because we already have state matching the gamedata, so we ignore the gamedata and don't order a new move.

Oh, here's an idea, don't ignore the 2nd gamedata if the bot isn't running?

roy7:
I need to stop second guessing the tests!

wonderingabout:
> I need to stop second guessing the tests!

haha
at least now it works xD

sanderland:
> I'm guessing why others never had this happen but @sanderland is, is maybe his server's connection to OGS is more laggy so he's generating these extra gamedata packets and we aren't.

Possibly combined with the fact that my bots are running a lot more games. 

>  I'm not sure the objection. Do you want to kill the bot earlier before we do the state setup instead of where it is now?

```  if (this.bot && gamedataChanged) {
                this.log("Killing bot because of gamedata change after bot was started");```

I want this at the top, and possibly `gamedataChanged` eliminated completely, just to simplify the logic

roy7:
@sanderland Reported this on discord.

> May 13 14:36:25   bot.js:148                     [14265] Stopping bot
> May 13 14:36:25   bot.js:148                     [14265] Attempting to send a command to dead bot: quit
> May 13 14:36:25   game.js:504                    [Game 23901707]            gamedata       B Zuzulu  [19x19]
> May 13 14:36:25   game.js:504                    [Game 23901707]            Game over.   Result: B+Can  W
> May 13 14:36:25   game.js:504                    [Game 23901707]            clock: {"game_id":23901707,"..}
> May 13 14:36:25   bot.js:148                     [14265] Bot exited
> May 13 14:36:25   game.js:504                    [Game 23901707]            Unable to react correctly - re-connect to trigger action based on game state.
> May 13 14:36:25   game.js:504                    [Game 23901707]            gamedata       B Zuzulu  [19x19]
> May 13 14:36:25   game.js:504                    [Game 23901707]            clock: {"game_id":23901707,"....}
> May 13 14:36:25   bot.js:148                     [12897] Killing process directly with a signal
> May 13 14:36:26   bot.js:148                     [12983] Killing process directly with a signal
> May 13 14:36:26   bot.js:148                     [13068] Killing process directly with a signal
> May 13 14:36:26   connection.js:664              #  disconnectFromGame 23901707
> May 13 14:36:26   game.js:504                    [Game 23901707]            Disconnecting from game.
> May 13 14:36:26   bot.js:148                     [13152] Killing process directly with a signal
> May 13 14:36:27   bot.js:148                     [13240] Killing process directly with a signal
> x10
>

roy7:
Can you paste more of the log or clarify what message was repeating 10x?

Also can you share logs from the 1-2 seconds prior? ScheduleRetry is 1000 ms delay, so it was probably called during 14:36:24 before the log starts.

roy7:
The "Killing process" messages are on a 5 second delay when bots exit. So that likely was related to something that happened 5 seconds before the log.

roy7:
We also want to stop processing the rest of the code if game is already finished, but be sure to call gameOver() first. Easy enough to adjust but @sanderland already put in https://github.com/online-go/gtp2ogs/pull/273 so we'll just go with his instead.

PR #271, opened 2020-05-12
Non square board size support

sanderland:
Important feature for 1x25 go


bot.js
```diff
@@ -629,11 +631,11 @@ function pretty_char2num(ch) {
     if (ch === ".") return -1;
     return "abcdefghjklmnopqrstuvwxyz".indexOf(ch.toLowerCase());
 }
-function move2gtpvertex(move, board_size) {
+function move2gtpvertex(move, bot, height) {
```
wonderingabout, line 634: here the `bot` parameter seems unused
something missing ?

sanderland: ah this is 'width', and is not used but here so the functions all look the same

bot.js
```diff
@@ -629,11 +631,11 @@ function pretty_char2num(ch) {
     if (ch === ".") return -1;
     return "abcdefghjklmnopqrstuvwxyz".indexOf(ch.toLowerCase());
 }
-function move2gtpvertex(move, board_size) {
+function move2gtpvertex(move, width, height) {
```
wonderingabout, line 634: why do you pass `width` even though you dont use it in the function?
same question as https://github.com/online-go/gtp2ogs/pull/271#discussion_r424265135

sanderland: I think it's a tad clearer and less likely to cause errors due to typos since all the previous code passed 'width' (and would silently cause errors if used like that).

wonderingabout: hmm i'd still not pass parameters i don't use, if other parts of the code are unclear it is those parts that should be made clearer, not adding extra confusion in a clean code i think

but i don't know, maybe @Dorus and @roy7 have better opinions

wonderingabout:
> Important feature for 1x25 go

you mean 25x1 go!!! :smile:

bot.js
```diff
@@ -487,7 +491,7 @@ class Bot {
                 if (!resign && !pass) {
                     if (move && move[0]) {
                         x = gtpchar2num(move[0]);
-                        y = state.width - parseInt(move.substr(1))
+                        y = state.height - parseInt(move.substr(1))
```
roy7, line 494: I'm curious why this was changed? Does it matter?

bot.js
roy7: I think it's best to pass both sizes consistently to all functions like @sanderland is doing even if some of the info isn't used internally by the function at this time

PR #273, opened 2020-05-15
If we get a duplicate gamedata and nothing has changed, ignore it, but better

sanderland:


game.js
```diff
@@ -39,50 +39,31 @@ class Game {
...
-
             // Only call game over handler if game really just finished.
             // For some reason we get connected to already finished games once in a while ...
-            if (gamedata.phase === 'finished' && prev_phase && gamedata.phase !== prev_phase)
-                this.gameOver();
+            if (gamedata.phase === 'finished') {
+                const prev_phase = (this.state ? this.state.phase : null);                
+                this.state = gamedata;                
+                if (prev_phase && gamedata.phase !== prev_phase) this.gameOver();
+                return; // ignore if we already handled it
```
roy7, line 48: The comment here doesn't seem to fit. Did you maybe mean to put this before the prior if() line?

I'm not sure which is more clean. How you do this now, or more like

if (gamedata.phase === 'finished') {
    if (this.state && gamedata.phase !== this.state.phase) {
        this.state = gamedata;
        this.gameOver();
    }
    return;
}

Just to shorten it, remove temp variable, and not assign to this.state if not needed. Not a bit deal.

roy7: Maybe it's better to assign every time though so this.state is always up to date. Even if we think we'll never need it today.

game.js
```diff
@@ -39,50 +39,31 @@ class Game {
...
 
             // If server has issues it might send us a new gamedata packet and not a move event. We could try to
             // check if we're missing a move and send it to bot out of gamedata. For now as a safe fallback just
             // restart the bot by killing it here if another gamedata comes in. There normally should only be one
             // before we process any moves, and makeMove() is where a new Bot is created.
             //
-            const gamedataChanged = (JSON.stringify(this.state) !== JSON.stringify(gamedata));
-
             if (this.bot && gamedataChanged) {
+
```
roy7, line 66: Intentional newline?

sanderland: no, not sure what all these are doing

game.js
sanderland: done

PR #295, opened 2020-05-22
hide jwt in debug, hide apikey and jwt only if they exist, keep track of notification to spot unexpected elements we receive from server

wonderingabout:
tested to work

edit: for now we'll just hide jwt, and see later how we want to modify debug information in the future

jwt hidden:
(just hide jwt and keep everything else in debug api data as it is)

jwt hidden

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -560,11 +560,6 @@ class Connection {
 function request(method, host, port, path, data) {
     return new Promise((resolve, reject) => {
         if (config.DEBUG) {
-            // Keeping a backup of old copy syntax just in case.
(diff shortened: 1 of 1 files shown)


wonderingabout:
as @Dorus said on discord, we should think more about how we want to modify debug information

so for now i rebased it to just hide jwt same as in the first screenshot

wonderingabout:
i added an extra enhancement: hide apikey and jwt only if they exist

old code would create a new apikey (and jwt) even if they didnt exist

wonderingabout:
i'm not sure if the tiny space gained in log is worth the extra code, should i close this pr?
what do you think?

Issue #294, opened 2020-05-22
"Killing bot because of gamedata change after bot was started" loop crash spotted


wonderingabout:
it seems we finally found a full and simple log to analyze the gamedata change loop crash!! :100: 

# [Download full log (5.5 MB)](https://drive.google.com/file/d/1xZjlsrEIK-1laXjwwj5zwLOEwVQsQWnw/view?usp=sharing)

first line where the bug is spotted is 16325

opening the log file with vs code, we can see that before line 16325 the "killing process directly with a signal" works normally, only once at game end

but starting from line 16325, we have 104 "killing bot because of gamedata change", from time 18:52:29 to time 18:52:44 the bot crash loops repeatedly,
until line 35474 (approximately 1000 lines per second)

@roy7 found on discord the issue and we are working on a fix,

extra screenshots (see yellow bars at the righest for ctrl+f results)

- 104 results for "killing bot because of gamedata change", most crowded at the same place

![gamedata change vs code](https://user-images.githubusercontent.com/38690718/82698656-77966480-9c6b-11ea-9ce1-279b525098b8.png)

- 219 results for "killing process directly with a signal", most abnormally matching the game data changes in blocks such as this one (see screenshot below too)

![vscode rightest](https://user-images.githubusercontent.com/38690718/82698637-6d746600-9c6b-11ea-95cb-cab98f5c5648.png)

Shell
```May 22 18:52:40   bot.js:148                     [52309] Killing process directly with a signal
May 22 18:52:40   bot.js:148                     [52320] Killing process directly with a signal
May 22 18:52:40   bot.js:148                     [52321] Killing process directly with a signal
May 22 18:52:40   bot.js:148                     [52332] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52335] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52346] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52347] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52358] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52372] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52383] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52394] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52405] Killing process directly with a signal
May 22 18:52:41   bot.js:148                     [52423] Killing process directly with a signal
[...]

PR #296, opened 2020-05-23
Delete gamedata.clock.now before processing gamedata packets

roy7:
This is to solve https://github.com/online-go/gtp2ogs/issues/294.

### DIFF
--- a/game.js
+++ b/game.js
@@ -126,6 +126,11 @@ class Game {
 
         this.socket.on(`game/${game_id}/clock`, (clock) => {
             if (!this.connected) return;
+
+            // Server has an issue that gamedata.clock.now will exist inconsistently. This will cause
+            // false positives for gamedata changes. We never use the field, so just remove it.
+            delete clock.now;
+
             if (config.DEBUG) this.log("clock:", JSON.stringify(clock));
 
             if ((config.nopause && !config.nopauseranked && !config.nopauseunranked)
--- a/game.js
+++ b/game.js
@@ -39,6 +39,10 @@ class Game {
         this.socket.on(`game/${game_id}/gamedata`, (gamedata) => {
             if (!this.connected) return;
 
+            // Server has an issue that gamedata.clock.now will exist inconsistently. This will cause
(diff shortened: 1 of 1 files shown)


wonderingabout:
maybe remove the commented-out line or comment something about it too?
https://github.com/online-go/gtp2ogs/blob/devel/bot.js#L188

roy7:
Since we don't update gamedata.clock from clock events, the commented out code is actually completely separate. I wrongly assumed we updated gamedata.clock when a clock event arrived. Oops. :) We could remove it since it's old and commented out but it's actually unrelated.

wonderingabout:
yes :)

i'll test this code for one week or something and tell you if i find any issue (can send all my logs at the end of the week ;)

testing branch is:
https://github.com/wonderingabout/gtp2ogs/tree/rejectnewtime-v2

wonderingabout:
similarly to this issue, it may also be a good idea to keep track of notifications we receive in debug because we never know if we don't receive any unexpected element from server
i am testing this in #295

wonderingabout:
i got a few clock.now elements errors again this hour, even though there was no more crashloop

see full log:
 # [Download Full Log (2.1 MB)](https://drive.google.com/file/d/1xZjlsrEIK-1laXjwwj5zwLOEwVQsQWnw/view?usp=sharing)

also i got 502 gateway errors on discord, various chromium websites, etc, so its possible few internet shortages and/or instability are involved in tonight's issue

wonderingabout:
this is from today's log (7748 and 7849)

clock.now is the only difference:

gamedata error again

roy7:
Oh I'm dumb. We don't store this.state.gamedata. this.state IS gamedata. So when we update this.state.clock we're changing our internal gamedata, and thus this.state.clock.now will appear from clock events after all.

wonderingabout:
great :smiley: 
i'll add your new change and continue the testing :1st_place_medal: 

and being dumb is a great quality for being a developer ;)

also i suggest we rapidly merge this if it doesnt break anything, it's a major achievement just on its own, and we can add further fixes later :)

PR #289, opened 2020-05-17
(tested but not exhaustively yet) add --rejectnewthreshold, refactor check_rejectnew into get_rejectnew_result

wonderingabout:
update 3: refactored check_rejectnew into get_rejectnew_result, not yet tested

update 2: tested to work, ready to merge :+1: 

- case 1: reject today works :+1: 
- case 2: reject tomorrow works too :1st_place_medal: 

this PR is a more modest attempt at #284 than #238

PR #297, opened 2020-05-23
(untested) tests don't detect exports from argv, declare them in tests, not at root

wonderingabout:
same as what was done for get_rejectnew_result in #289


wonderingabout:
this is totally wrong way of doing it, will be properly implemented in #287

PR #301, opened 2020-05-24
Ignore auto_score from server, can be inconsistent

roy7:


wonderingabout:
i'll add this to my testing in https://github.com/wonderingabout/gtp2ogs/tree/rejectnewtime-v2

PR #303, opened 2020-05-27
Add hide() unhide() setHidden() helper functions

roy7:
For future use by the various clock/time rejection stuff. We could set timers to auto hide an unhide or at least hide if we're shut off for good, so we don't appear in server drop down any more.

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -555,6 +555,15 @@ class Connection {
         clearInterval(this.notification_connect_interval);
         clearInterval(this.corr_queue_interval);
     }
+    setHidden(setting) {
+        socket.emit('bot/hidden', !!setting);
+    }
+    hide() {
+        socket.emit('bot/hidden', true);
+    }
+    unhide() {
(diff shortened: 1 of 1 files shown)


wonderingabout:
Very great!!!
first thing we could apply it would be rejectnew to begin with :)

wonderingabout:
it seems the travis isnt working, but this may just mean `socket` needs to be defined in tests

i'll test this on my branch and tell you if it works, if it does then we can fix the test :)

Dorus:
Dont forget to use the new function on line 81 https://github.com/online-go/gtp2ogs/pull/303/files#diff-0d74c7455b8cbe5be6f644690ecb1b8aL81

Ok from me otherwise

roy7:
@Dorus Is it just redundant though and maybe I'll remove setHidden completely?

wonderingabout:
i tested this on beta, we successfully trigger hiding and unhiding our bot, it works good :+1:

PR #316, opened 2020-06-13
Set defaults to ranked if only unranked is set and vice versa.

Dorus:


roy7:
Are we planning to move away from general eventually? If so should `argv[general] = family.default;` instead say `argv[unranked] = family.default;  argv[ranked] = family.default;`?

PR #320, opened 2020-06-15
(wip) Exports option ranked unranked

wonderingabout:
(unfinished)

sending it as a pull request so that we can compare and track changes more easily

all files were still not changed to the new syntax, so it is expected that tests currently fail


wonderingabout:
i forgot to comment in commit notes of last commit:
by not exporting the group options's argvs (ex: `"9,13,19"` for `--boardsizes`), we are leaking less uneeded information from the config

we can also directly check if notif is in config.ranked.boardsizes's keys, so we don't need to check if arg is defined here neither

the only place where the group's argv is actually used is for parsing boardsizes out of argv `"9,13,19"`, but even in that case we can get all the keys we want by doing `Object.keys(config.boardsizes.allowed)`, so argv is redundant in connection.js
Also, if we modify our config parsing (ex: by adding an increment and range operator ex: `"9,13:19"` -> `"9,13,14,15,16,17,18,19"`, then our abstract boardsize function will be unable to correctly parse `13:19x13:19`, however if it worked on an array of `Object.keys(config.boardsizes.allowed)` allowed boardsizes one by one, then it would work automatically and well

so all in all, argv is redundant and uneeded to export from config.js

wonderingabout:
big advances were made today!
now we successfully export ranked unranked defaults in test.js and all testing files, and basic unit testing has been added for most options in connection.js

for the minmax options, minmaxrank unit testing was already successully added using the new config.ranked and config.unranked syntax

this is a lot closer from being finished now :tada:

roy7:
@Dorus Think this is worth leaving open for someone to finish down the road or just close it?

Dorus:
Too many conflicts here and no Meta anyore to fix them. There are some good ideas in here but also a lot of changes and it's a bit messy. I say we close it, if we ever want to redo this, the code will still be there for us to look at.

PR #323, opened 2020-06-17
Report games needed for min games.

Dorus:


wonderingabout:
this is good to merge
it will help our bot's reliable users who are confused by the mingames message
our focus should be a good experience for reliable users, not a bad experience for harmful ones
:100:

PR #324, opened 2020-06-18
Fix mingamesplayed documentation

wonderingabout:
related to #323


wonderingabout:
reduced maxmimum value recommended from 20 to 10

Dorus:
This should be ok to marge, since it's only docs.

wonderingabout:
due to a change on OGS, this is not relevant anymore,
see #324

PR #331, opened 2020-06-23
minor formatting: skip lines in test.js

wonderingabout:
added lines skip to match this block of code:
<https://github.com/online-go/gtp2ogs/blob/devel/test/test.js#L741-L769>

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -797,22 +797,31 @@ describe("Pv should work", () => {
     function testPv(pvCode, fileName, chatBody) {
         stub_console();
         sinon.useFakeTimers();
+
         let fake_socket = new FakeSocket();
         let fake_api = new FakeAPI();
         fake_api.request({ path: '/foo' }, () => { });
         sinon.stub(https, 'request').callsFake(fake_api.request);
+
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -797,22 +797,31 @@ describe("Pv should work", () => {
...
         let conn = new connection.Connection(() => { return fake_socket; }, config);
+
         const game = sinon.spy();
         game.sendChat = sinon.spy();
         game.processing = true;
         game.state = { width: 19, moves: { length: 2 } };
+
         config.ogspv = pvCode;
+
         new Bot(conn, game, config.bot_command);
```
wonderingabout, line 818: is something missing in this line?

should it be, for example, something such as:

```Js
        const bot = new Bot(conn, game, config.bot_command);
```

in <https://github.com/online-go/gtp2ogs/blob/devel/test/test.js#L757>

Dorus: not really, we're not going to use the bot class from here. It's used trough the attached bot fake.

Dorus:
This one is ok to merge @roy7

PR #332, opened 2020-06-24
add unit testing for config.js and partially getArgv.js, set logfile console before entering config

wonderingabout:
current status:
- ready to merge
- could add custom errors as @Dorus explained in https://github.com/online-go/gtp2ogs/pull/332#discussion_r464049181


wonderingabout:
i added stub_console, but console output is still shown in tests

wonderingabout:
i just rebased this on top of latest #333 , is it good to merge ?

wonderingabout:
hiding console in tests finally works for config too!!!!!!!!

console stub is successful

merge conflicts are fixed, but after i pulled pr12 from @Dorus this is the error message i get:

current error

wonderingabout:
this is the current error after running `node --trace-warnings` as asked by node

current error traced

wonderingabout:
pull from latest devel automatically worked, then i rebased this
i also tested it on beta ogs to work, and all tests are green too

i have 3 questions about this pr :

1) the problem of circular dependencies is still existing, even though config is correctly exported to connection, and this branch works well on beta ogs:

circular dependencies

2) should we remove all defaults from yargs in getArgv.js (for example moving them to constants.js and having a function that sets them if argv is === undefined), and set them all in config.updateFromArgv same as the rankedUnrankedOptions?
This would allow us to get all config defaults natively and faithfully replicate the config behaviour on real ogs

3) the 2) would also allow me to write more tests without manually inputting default one by one (for example minmaintimeblitz, etc.), so more generally should i write a few more options tests (especially pv for example) ?

Dorus:
Re2: Does Yarg show the defaults in the help msg? If so you can put them in a constant, but still need to use yarg to loop over them and default them.

wonderingabout:
yes, and this is so nice wow

i added `yargs.showHelp();` in getArgv.js just before the return yargs.argv, and this is the result:
https://pastebin.com/rYqycmsZ

wonderingabout:
so put defaults in constants.js, import them in getArgv, and set them using the constant rather than hard code the default :trophy: 

if we do that, how would we set our defaults in config.test.js?
we could in test/getNewArgvReset add a function that sets them one by one, but would that be a repetition then?

i'll test this this evening and keep you updated :)
this is so fun :)

wonderingabout:
but we spent a lot of time on gtp2ogs recently, i wouldnt want us to be too fast and exhaust ourselves

so we can review it this weekend or some future time ;)
i'll try a few things and keep you updated :1st_place_medal:

wonderingabout:
interesting findings, about the `Warning: Accessing non-existent property 'DEBUG' of module exports inside circular dependency` error, i found it started from node 14+, so if you're still on version 13- you will not see it @Dorus 

also, even on node 14+, the issue does not appear on devel, so it has to be something we do specifically on this branch that creates this issue, i'll start from devel again and add features from this branch one by one until we find the mistake

wonderingabout:
i found the error causing circular dependency, it is at line 13 of config.js

```Js
const { console } = require('./console');
```

removing this line fixes the issue

wonderingabout:
after more investigating, our custom console.js has all this and causes the circular dependency error:

`{"_stdout":{"connecting":false,"_hadError":false,"_parent":null,"_host":null,"_readableState":{"objectMode":false,"highWaterMark":16384,"buffer":{"head":null,"tail":null,"length":0},"length":0,"pipes":[],"flowing":null,"ended":false,"endEmitted":false,"reading":false,"sync":true,"needReadable":false,"emittedReadable":false,"readableListening":false,"resumeScheduled":false,"errorEmitted":false,"emitClose":false,"autoDestroy":false,"destroyed":false,"errored":false,"closed":false,"closeEmitted":false,"defaultEncoding":"utf8","awaitDrainWriters":null,"multiAwaitDrain":false,"readingMore":false,"decoder":null,"encoding":null,"readable":false},"_events":{},"_eventsCount":1,"_writableState":{"objectMode":false,"highWaterMark":16384,"finalCalled":false,"needDrain":false,"ending":false,"ended":false,"finished":false,"destroyed":false,"decodeStrings":false,"defaultEncoding":"utf8","length":0,"writing":false,"corked":0,"sync":true,"bufferProcessing":false,"writecb":null,"writelen":0,"afterWriteTickInfo":null,"buffered":[],"bufferedIndex":0,"allBuffers":true,"allNoop":true,"pendingcb":0,"prefinished":false,"errorEmitted":false,"emitClose":false,"autoDestroy":false,"errored":false,"closed":false},"allowHalfOpen":false,"_sockname":null,"_pendingData":null,"_pendingEncoding":"","server":null,"_server":null,"columns":118,"rows":46,"_type":"tty","fd":1,"_isStdio":true},"_stderr":{"connecting":false,"_hadError":false,"_parent":null,"_host":null,"_readableState":{"objectMode":false,"highWaterMark":16384,"buffer":{"head":null,"tail":null,"length":0},"length":0,"pipes":[],"flowing":null,"ended":false,"endEmitted":false,"reading":false,"sync":true,"needReadable":false,"emittedReadable":false,"readableListening":false,"resumeScheduled":false,"errorEmitted":false,"emitClose":false,"autoDestroy":false,"destroyed":false,"errored":false,"closed":false,"closeEmitted":false,"defaultEncoding":"utf8","awaitDrainWriters":null,"multiAwaitDrain":false,"readingMore":false,"decoder":null,"encoding":null,"readable":false},"_events":{},"_eventsCount":1,"_writableState":{"objectMode":false,"highWaterMark":16384,"finalCalled":false,"needDrain":false,"ending":false,"ended":false,"finished":false,"destroyed":false,"decodeStrings":false,"defaultEncoding":"utf8","length":0,"writing":false,"corked":0,"sync":true,"bufferProcessing":false,"writecb":null,"writelen":0,"afterWriteTickInfo":null,"buffered":[],"bufferedIndex":0,"allBuffers":true,"allNoop":true,"pendingcb":0,"prefinished":false,"errorEmitted":false,"emitClose":false,"autoDestroy":false,"errored":false,"closed":false},"allowHalfOpen":false,"_sockname":null,"_pendingData":null,"_pendingEncoding":"","server":null,"_server":null,"columns":118,"rows":46,"_type":"tty","fd":2,"_isStdio":true},"_ignoreErrors":true,"_times":{}}`

but the native node console has only this and doesnt cause the issue
`{}`

wonderingabout:
ah i found where is the issue!!!!

it is because in console.js () we reference:

- config.DEBUG (https://github.com/online-go/gtp2ogs/blob/devel/console.js#L9)
- and config.logfile (https://github.com/online-go/gtp2ogs/blob/devel/console.js#L27)

 but these are not yet assigned when we import console in config.js (config is not yet exported)!!!

so we could remove mentions of config in console.js, or create another custom console specially for config.js, or keep using the native node console as we did so far, but then we'd lose timestamp and other formatting

wonderingabout:
so now we're back to the issue on how to stub console on tests,

possibly we could create our own console specifically for config, so stubbing it would stub console logging and greeting, without stubbing the native console that displays the tests pass/fail reult

wonderingabout:
it seems after all we may want to run console.js based on argv.logfile and argv.debug, not based on config

wonderingabout:
fixed!

now console.js can be used in config.js too!!!!

yargs argv add argv

i need to remove the JSON.stringify debug line,
and then also modify all our tests to create an argv just like in real gtp2ogs, so that console.js can use this argv too in tests and so that we can stub our console.js without stubbing the node console that displays tests pass/fail result

wonderingabout:
but there are still some problems with this approach so i'm not sure it's a good idea to keep it

it could be a problem for example that everytime we require console, we get a new argv, so if `--logfile` is used it may mean creating a new file with a new `new Date().toISOString()`, so we may instead want to export argv once and never touch it again, same as what we do for config, but it would create needless duplication, as we dont need argv outside of config (and console.js now)

wonderingabout:
we now can test if errors are thrown in config!!!

i ran a few tests and some googling until i found the committed solution :100: 

TODO is to add testing for logfile and ogsPVAIs at least, especially for logfile after we merge (#354 )

here is an example of one of my unsuccessful attemps that shows we do listen to console throwing errors!! :tada: 

vscode assert throw


TODO 2 is also to fix the circular dependency that we're still getting now that i reverted to a few commits ago

wonderingabout:
imported all changes from #332 and added more tweaks and unit testing for logfile !!! :trophy: :trophy: :trophy: :trophy:

wonderingabout:
current TODO for this PR is to decide how we want to fix the circular dependencies issue

- get logfile and debug info from argv?
- create a console2 specially for config.js that did not check and export config.DEBUG and config.logfile yet but uses them

another idea? :thinking:

wonderingabout:
another TODO is to reference argv defaults in yargs in getArgv.js, and reference them by variable address, so that we can important constants.js in test.config.js and run other tests without repeating them

also it may be a good idea to add argv in all other test files, then run a config.updateFromArgv(argv), so that we get a config similar to the real config (with defaults and all other behaviours) in our tests

wonderingabout:
interesting, my node 14+ testing environment already detected the circular dependency error, but it was too small to notice it, maybe we should add `--trace-warnings` to our tests as is suggested :thinking: 

but it's great news that our test could even detect that, thanks to @windo ! :100:

wonderingabout:
while writing testing for general ranked unranked defaults, i noticed our args are not even parsed correctly!!!

--minmaintimeblitzranked 11 will use default 15 for minmaintimeblitzranked

currently fixing this then writing more tests for GRU rules

wonderingabout:
fixed!!!

the fix was so simple but it was easy to miss in all this code and without testing!

i think we should make a new release after this PR is merged, it fixes logfile crash and ranked unranked args uncorrectly exported, which are both quite important issues

wonderingabout:
move exports.check_rejectnew function in config.js to a local check_rejectnew function in connection.js!!

also added unit testing for rejectnew and rejectnewfile

and tested to work on beta

check_rejectnew is in connection

it was very interesting to learn how `fs` takes relative and absolute paths for files it uses

wonderingabout:
this PR is getting too long, so i should make smaller PRs for some features that are ready to merge (and unit testing for these features will be added in this PR)

wonderingabout:
here for logfile in config.test.js we could also check if a file was created at specified location using fs.existsSync same as what was already done in check_rejectnew

wonderingabout:
successfully merged 360 into 332

test/module_loading/getNewArgvWithArguments.js
```diff
@@ -0,0 +1,15 @@
+const { rootOptionsDefaults } = require('../../constants');
+
+function getNewArgvWithArguments() {
+    return {
+        username: 'testbot',
```
wonderingabout, line 5: this is currently a repetition of /test/assignConfigArguments.js

should we always import defaults from same source (for example from the real getArgv.js as in #358 , or by creating a common shared /test/testDefaults.js where testing's username, apikey, etc. are declared once and required in all modules that use it?

also, currently we dont import all defaults from the real getArgv, only rejectnewmsg, current tests, especially in checkChallenge.test.js should be slightly updated to match defaults (for example it's unlikely to specifically declare an option as undefined if it has a default (ex: `config.minmaintimeblitz = undefined` which tests `--minmaintimeblitz undefined`)

this may be done in a separate PR rather

config.js
```diff
@@ -8,6 +8,10 @@ const { getOptionName } = require('./options/getOptionName');
 const { getRankedUnranked } = require('./options/getRankedUnranked');
 const { getRankedUnrankedUnderscored } = require('./options/getRankedUnrankedUnderscored');
 
+// TODO fix circular dependency
```
Dorus, line 11: todo?

config.js
```diff
@@ -19,7 +23,8 @@ exports.updateFromArgv = function(argv) {
     const debugStatus = argv.debug ? "DEBUG: ON\n" : "";
     console.log(`\ngtp2ogs version 6.0.1\n--------------------\n- For changelog or latest devel updates, please visit https://github.com/online-go/gtp2ogs/tree/devel\n${debugStatus}`);
 
-    // B - test unsupported argv
+    // B - throw errors
+    // all errors should be tested and thrown before we start exporting
```
Dorus, line 27: I'm not quite sure what this comment means. It's better to describe what we're going to do here, or not put a comment at all.

config.js
```diff
@@ -105,51 +103,61 @@ function exportInitialConfigUnderscored(rankedUnrankedUnderscored) {
...
+
 function testRankedUnrankedOptions(rankedUnrankedOptions, argv) {
     for (const option of rankedUnrankedOptions) {
         const [general, ranked, unranked] = getArgNamesGRU(option.name);
         
         // check undefined specifically to handle valid values such as 0 or null which are tested false
         if (argv[general] !== undefined) {
             if (argv[ranked] !== undefined) {
-                throw `Cannot use --${general} and --${ranked} at the same time.\nFor ranked games, use either --${general} or --${ranked}, or do not use anything to switch to default settings for this option.`;
+                throw new Error(`Cannot use --${general} and --${ranked} at the same time.\nFor ranked games, use either --${general} or --${ranked}, or do not use anything to switch to default settings for this option.`);
```
Dorus, line 121: If we're going to throw error objects, it might be better to extewnd the error object, and later when we catche it, use `instanceof` to check if it's our own custom error.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types

config.js
wonderingabout: currently there is an issue with circular dependencies (https://github.com/online-go/gtp2ogs/pull/332#issuecomment-654633308), it was a reminder to fix it before merging
i should delete this after this is fixed, or possibly now too

config.js
wonderingabout: this comment is related to pr #362 , which i dont think the idea is too good, the idea was to avoid the escape of `""` minmaxranks that dont throw an error even though they are invalid (a much simpler fix would have been to just do `!== undefined` instead of `if (argvRank)` in processRankExport, but i thought to generalize it (even though again this idea seems quite bad in the end, i may revert it to devel and close #362 , what do you think?)

config.js
wonderingabout: oh very interesting, it is where i currently stopped at the js lesson
i'll try it then 💯

config.js
```diff
@@ -253,14 +258,16 @@ function parseRank(arg) {
                 return (36 + parseInt(results[1]));
             }
         } else {
-            throw `error: could not parse rank -${arg}-`;
+            throw new Error(`error: could not parse rank -${arg}-`);
         }
+    } else {
```
wonderingabout, line 263: this is a much easier fix to the empty string escape

wonderingabout:
i rebased this on top of latest devel in one commit :1st_place_medal:

config.js
Dorus: This is fixed, guess i forgot to remove this? 😅

wonderingabout: yes i forgot it too actually :sweat_smile: :sweat_smile: :sweat_smile:

wonderingabout:
but after merging @Dorus 's fix, it fixes circular dependency error, but another problem is that the logfile is no longer created :open_mouth: 

more than just fixing this, i would like to take this opportunity to write unit testing that tests we create a logfile (possibly using mock-fs https://www.npmjs.com/package/mock-fs)

wonderingabout:
this commit https://github.com/online-go/gtp2ogs/pull/332/commits/befd71b267971144e0159016cf226b4754307147 adds mock-fs and testing for console.js

current issues are:
- cache problem, console.test.js tests work individually, but when all ran at the same time mocked directory content is not reset, so after test 1 total number of files is 2 (expected 1), and after test 2 total number of test is 1 (expected 0)
- conflict between fs in other test files: game.test.js and pv.test.js use for some reason the mocked fs even after `mock.restore()`, thats why it cant find pv samples in non our fake directories (these are in the real directories)

other issues unrelated to this commit:
- after https://github.com/online-go/gtp2ogs/pull/332/commits/6bbd216781bd22e3e191adf6b8418252c100f505 's merge, some tests are not stubbed anymore in config.test.js (this should be because of the additional console declaration in setLogfile in config.js)

wonderingabout:
game works end to end on beta ogs
a logfile is successfully created and all console is logged into file
all tests pass
config.test.js successfully stubs console for all tests

this should be good to merge, but may require code optimization :1st_place_medal:

wonderingabout:
i rebased all the latest changes in 1 commit after latest pr merge from @Dorus : https://github.com/online-go/gtp2ogs/pull/332/commits/87585243fb6d87cd9fa4979ca9ceb8c3083db878

roy7:
@Dorus Here's final pending one if you want to review sometime, before anoek sneaks in to merge it out of nowhere. ;)

anoek:
Closing since things have diverged a bit, feel free to reopen if there's eventual interest in picking this back up

PR #333, opened 2020-06-24
split config.updateFromArgv() into getArgv() and config.updateFromArgv(argv), add constants.js

wonderingabout:
tested to work on beta and on real ogs

see questions here:
https://github.com/online-go/gtp2ogs/pull/333#issuecomment-649031885


wonderingabout:
i have a few questions about this PR:

0)
getArgv.js (function) or argv.js (class) ?

1)
how to remove ogsPvAIs double assignment repetiton in getArv.js https://github.com/online-go/gtp2ogs/pull/333/files#diff-a13be434fa77bb7162b4b5a39159bcecR1 and config.js https://github.com/online-go/gtp2ogs/pull/333/files#diff-e22b50e740e102c8de5e858dbcd7753dR38

i thought of not passing any parameter, ogsPvAIs would be guessed from the scope but i didnt like it as it's less clear and more error prone,
also i thought of making a gloabal declaration of ogsPvAIs in gtp2ogs.js but it seemed way too overkill and unclean for something we only use in config,
i also thought about exporting it but similarly it also seems uneeded as we dont use it outside of config

so for now i just copy pasted it twice,
what could we do?

2)
about abstraction: so we just get about argv in getArgv.js then modify it and test it as much as we want, or do we also do other tasks (ex: testBotCommand from #314) and/or set defaults for ranked/unranked (and then we may need to pass rankedunranked options twice too)

wonderingabout:
re on abstraction, i gave it some thought and i think we may want to create an argv.js file that does several operations on argv

it would first get argv from optimist, then set rankedunranked defaults, test bot command is valid, tests dropped argv, etc., then finally if argv is valid we export it to config

and for config we would only run config.updateFromArgv(argv);

what do you think?

getArgv.js
```diff
@@ -0,0 +1,179 @@
...
+        .describe('minperiodtimecorrranked', 'Minimum seconds of period time for correspondence ranked games')
+        .describe('maxperiodtimecorrranked', 'Maximum seconds of period time for correspondence ranked games')
+        .describe('minperiodtimecorrunranked', 'Minimum seconds of period time for correspondence unranked games')
+        .describe('maxperiodtimecorrunranked', 'Maximum seconds of period time for correspondence unranked games')
+    ;
+
+    return optimist.argv;
+}
+
+exports.getArgv = getArgv;
```
Dorus, line 179: newline

getArgv.js
```diff
@@ -0,0 +1,179 @@
+function getArgv() {
+    const ogsPvAIs = ["LeelaZero", "Sai", "KataGo", "PhoenixGo", "Leela"];
```
Dorus, line 2: We do not want to duplicate this. We could make a const.js files or something like that that exports common constants like this.

gtp2ogs.js
```diff
@@ -8,8 +8,12 @@ process.title = 'gtp2ogs';
 
 // Do this before importing anything else in case the other modules use config.
 const io = require('socket.io-client');
+
+const { getArgv } = require('./getArgv');
+const argv = getArgv();
+
 const config = require('./config');
-config.updateFromArgv();
```
Dorus, line 12: I'm thinking this should be passed to the constructor of the config class, but i think that's too much of a refactor right now.

getArgv.js
wonderingabout: added :+1:

getArgv.js
wonderingabout: yes, duplication is bad

it could also be useful for other parts of the code where we reuse and redeclare shared constants

gtp2ogs.js
wonderingabout: no it may be a good opportunity to do it actually, but i dont know if we pass argv in the constructor and we do `this.argv = argv`, then would argv be exported in config.argv ? (which would be uneeded)

wonderingabout:
tested to work on beta ogs

config.js
```diff
@@ -10,6 +10,9 @@ const { getOptionName } = require('./utils/getOptionName');
 const { getRankedUnranked } = require('./utils/getRankedUnranked');
 const { getRankedUnrankedUnderscored } = require('./utils/getRankedUnrankedUnderscored');
 
+const ogsPvAIs = require('./constants').constants.ogsPvAIs;
+const rankedUnrankedOptions = require('./constants').constants.rankedUnrankedOptions;
```
Dorus, line 14: Can we not do

```Js
const { ogsPvAIs, rankedUnrankedOptions } = require('./constants').constants
```

Also. Can we not remove `constants` in there and make it use `require('./constants')` directly?

wonderingabout: yes, that would remove a repetition :1st_place_medal: 

i rebased on top of latest devel, and also moved droppedOptions to constants.js :100:

wonderingabout: tested to work on beta ogs

wonderingabout: tested to work on beta ogs :+1:

constants.js
```diff
@@ -0,0 +1,89 @@
...
+    { name: "maxperiodtimelive", default: 120 }, // 2 minutes
+    { name: "maxperiodtimecorr", default: 259200 } // 3 days
+];
+
+function getBLCString(optionName, rankedUnranked) {
+    return `${optionName}blitz${rankedUnranked}, --${optionName}live${rankedUnranked} `
+           + `and/or --${optionName}corr${rankedUnranked}`;
+}
+
+exports.constants = { droppedOptions, ogsPvAIs, rankedUnrankedOptions };
```
Dorus, line 89: can we not just drop the `constants` and just do
```Js
exports = { droppedOptions, ogsPvAIs, rankedUnrankedOptions };
```
and then also remove `constants` when we import?

wonderingabout:
i just retested on beta and this is ready to merge
config is imported successfully, game works end to end

constants.js
wonderingabout: ah i see, but then the exports become a bit messier no?

considering we use these only in config.js and getArgv.js, i find it clearer to wrap them all in the same module no ?

similar to the issue we had in #342 when we wanted to create a sub directory :sweat_smile:

wonderingabout:
current error after test commit:

current error

wonderingabout:
a few screenshots for info, in constants.js:

exports works v1

exports doesnt work v1

exports v3

exports v4

Dorus:
Oops i tried a few more things, the correct way to export all at once is
```Js
module.exports = { droppedOptions, ogsPvAIs, rankedUnrankedOptions };
```
Alternative you can also do 
```
exports.ogsPvAIs = [...]
```

PR #336, opened 2020-06-28
add sanity check: non-correspondence challenges cannot have pausing on weekends

wonderingabout:
edit2: i just rebased this, ready to merge

this fixes the issue @sanderland noticed in https://online-go.com/game/view/24994942, where live challenges could strangely have pausing on weekends

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -331,11 +331,19 @@ class Connection {
             return { reject: true, msg };
         }
 
-        // Sanity check: OGS enforces rules to be chinese regardless of user's choice.
+        // Sometimes server sends us live challenges with pauses on weekends enabled.
+        if (notification.time_control.pause_on_weekends && notification.time_control.speed !== "correspondence") {
+            err(`Unhandled pause on weekends in non-correspondence challenge (${notification.time_control.speed}).`);
+            const msg = `There was an unexpected error: your ${notification.time_control.speed} challenge`
+                        + ` has pause on weekends, but this is only possible for correspondence games`
+                        + `, please try again.`;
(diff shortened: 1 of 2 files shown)


connection.js
```diff
@@ -331,11 +331,18 @@ class Connection {
             return { reject: true, msg };
         }
 
-        // Sanity check: OGS enforces rules to be chinese regardless of user's choice.
+        // Sometimes server sends us live games with pausing on weekends.
+        if (notification.time_control.pause_on_weekends && notification.time_control.speed !== "correspondence") {
+            err(`No pausing on weekends in ${notification.time_control.speed}.`);
+            const msg = `Pausing on weekends is not possible in ${notification.time_control.speed}`
+                        + `, only for correspondence games, please contact my bot admin.`;
```
sanderland, line 338: why do they need to contact anyone?

wonderingabout: ah i just copied the same message for other sanity checks errors 😅 

it's true that we could just suggest them the correct fix to the error
but here in this case there should not even be a possibility to use pauses on weekends for live games, so either it is a bug, or user could have found an exploit
problem is that if this is a bug (so optin can't be disabled) and if botadmin uses a nopauseonweekends options, then it would permanently exclude all users from live games, which should be reported in that case because it is a bug

wonderingabout: but for other sanity checks knownSpeeds or knownTimecontrols, it is uneeded to contact botadmin and user could just try sending challenge again, which should fix the server very rare error

wonderingabout:
fixed according to @Dorus 's review :+1:

PR #338, opened 2020-06-29
move tests shared functions and classes to /tests/utils

wonderingabout:
edit: some of these will be shared with config.test.js in #332 , so they are move to utils in anticipation to alleviate the review later


wonderingabout:
is the idea of this pr good?

if yes, i can already implement it in #320

Dorus:
This one looks good. Can be merged @roy7

PR #340, opened 2020-07-04
update package-lock.json to 6.0.1

wonderingabout:
removed node_modules folder then ran `npm i` from latest devel

edit: tested to work on beta, ready to merge


Dorus:
Did you also run things like `npm audit fix` and update outdated packages? Might be a good thing to add here.

wonderingabout:
ah i didnt do it because i dont know about these things, i'll try it! :100:

Dorus:
Just make sure you test things still work after updating. Package-lock is made specifically to avoid breaking things because of random updates, but only update when developers request it so they can test if nothing randomly broke.

wonderingabout:
this should be merged after #342 as it needs some testing before merging, and is independent from the other important changes

also we should merge this pr before making a new 6.0.1 release

wonderingabout:
so far it looks like this, didnt run any audit yet

wonderingabout:
should redo it properly lol in another pr

PR #341, opened 2020-07-04
fix missing files pv.js and utils/** in package.json files for npm install

wonderingabout:
edit: this is also done in #333 where we also add constants.js to list of gtp2ogs files

quesiton:
should we also add .js test files here? (test.js and checkChallenge.test.js)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "connection.js",
     "console.js",
     "game.js",
-    "gtp2ogs.js"
+    "gtp2ogs.js",
+    "pv.js"
   ],
   "bin": {
     "gtp2ogs": "gtp2ogs.js"
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "console.js",
     "game.js",
     "gtp2ogs.js",
-    "pv.js"
+    "pv.js",
+    "/utils/**"
   ],
   "bin": {
     "gtp2ogs": "gtp2ogs.js"


Dorus:
Yes we need this. Good find!!

We messed up our gtp2ogs release royally, so stupid nobody complained about this so far :(

Please also add `utils/**` to this list.

wonderingabout:
sure, i guess it would be good to make a new release, either 6.0.1 or 6.1.0 i guess

so i just add `utils/**` ?

Dorus:
6.0.1 so it's easy to update for people.

Actually we might need to decide. Did we deprecate any command line option since the last release? If so we might want 6.1.0. In that case we could somewhat decide to also fork the old 6.0.0 release, fix the pv file and other missing files, and release bot 6.0.1 and 6.1.0 later.

wonderingabout:
i just committed the fix

for version name we didnt much commits since 6.0.0, we dropped fakerank but it's a thing none was using as far as i know, the rest is minor ;)

https://github.com/online-go/gtp2ogs/compare/6.0.0...devel

wonderingabout:
the suggestion to fork 6.0.0 seems a bit overkill, it's been only a few weeks since 6.0.0 release, and few people use gtp2ogs in general, so we may just drop support for old version and just release a bugfix 6.0.1 :thinking:

Dorus:
Thinking about this for 5 more seconds, yes 6.1.0 makes zero sense as the current 6.0.0 release is so broken we have zero reason to care about compatibility. We can just wack on 6.0.1 with any number of updates and fixes :)

Dorus:
@wonderingabout We should increase version number in this pr, or, we could close this one, increase version number in #342 since that one has everything this one has too.

wonderingabout:
> Thinking about this for 5 more seconds, yes 6.1.0 makes zero sense as the current 6.0.0 release is so broken we have zero reason to care about compatibility. We can just wack on 6.0.1 with any number of updates and fixes :)

yes i agree!! :medal_military: 



> Thinking about this for 5 more seconds, yes 6.1.0 makes zero sense as the current 6.0.0 release is so broken we have zero reason to care about compatibility. We can just wack on 6.0.1 with any number of updates and fixes :)

yes i agree!! :medal_military: 

> @wonderingabout We should increase version number in this pr, or, we could close this one, increase version number in #342 since that one has everything this one has too.

if we merge 642 latyer, then we can increase it there
i like that this pr commits specifically the fix, it's a good way to highlight that there was an issue

wonderingabout:
ah nvm it adds complications to merge this lol, better close it

PR #344, opened 2020-07-04
update package-lock.json to 6.0.1

wonderingabout:
this is the continuation of #340 with a cleaner base

i need to run an audit now as @Dorus explained in https://github.com/online-go/gtp2ogs/pull/340#issuecomment-653809523

since it is quite late today, i'll do it next time


wonderingabout:
result of npm audit fix

audit fix result

PR #345, opened 2020-07-05
drop optimist, use yargs instead

wonderingabout:
this PR is on top of #333 as it makes it a lot easier

i followed this documentation and could very easily implement yargs:
https://github.com/yargs/yargs/blob/master/docs/examples.md

as both optimist and yargs are "pirate themed" (:sweat_smile:), i guess they have a lot of architechture in common

see last commit for changes specific to this PR:
https://github.com/online-go/gtp2ogs/pull/345/commits/84d0397395368b96b1306299e4ca41f0f64d6d49

i also tested it to work on beta ogs

### DIFF
--- a/getArgv.js
+++ b/getArgv.js
@@ -2,7 +2,7 @@ const { ogsPvAIs } = require('./constants');
 
 function getArgv() {
 
-    const optimist = require("optimist")
+    const yargs = require("yargs")
         // 1) ROOT OPTIONS
         .usage("Usage: $0 --username <bot-username> --apikey <apikey> [gtp2ogs arguments] -- botcommand [bot arguments]")
         .demand('username')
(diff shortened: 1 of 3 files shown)


wonderingabout:
config is imported successfully, and a game works end to end:

yargs v1

yargs v2

wonderingabout:
fixed merge conflicts, tested on beta and it works fine, there isnt the same error as in #332 of circular dependencies

config.js
```diff
@@ -37,6 +37,9 @@ exports.allowed_timecontrols_ranked = {};
 exports.allowed_timecontrols_unranked = {};
 
 exports.updateFromArgv = function(argv) {
+    // TODO remove this debug line later after we finish testing yargs
```
Dorus, line 40: Do the todo, then we can merge this.

PR #353, opened 2020-07-07
minor fixes

wonderingabout:


test/e2e.test.js
```diff
@@ -18,8 +18,8 @@ const config = getNewConfig();
 const connection = require('../connection');
 const { console } = require('../console');
 
-config.timeout = 0; // needed for test.js
```
Dorus, line 21: This is e2e.test.js. We should rather delete this comment.

wonderingabout: yes,
it was a memento to remember @windo originally put in in config.js in https://github.com/online-go/gtp2ogs/commit/f5efe7b4b03c7e868fa15fa11d5773f90f533d37#diff-c1129c8b045390789fa8ff62f2c6b4a9R23-R24, while it seemed to be a duplicate that was needed only for test.js at that time

now we know this is not needed in config.js, so we can remove the comment indeed

Dorus:
This can be merged

PR #354, opened 2020-07-10
fix invalid characters in logfile filename (fixes #337)

### LINKED ISSUE #337: Client got suspended
Sometimes got this error message
```
Jun 29 22:06:06 !!!!! connection.js:706              #  520 - <html>
<head><title>520 Origin Error</title></head>
<body bgcolor="white">
<center><h1>520 Origin Error</h1></center>
<hr><center>cloudflare-nginx</center>
</body>
</html>

Jun 29 22:06:06 !!!!! connection.js:707              Error
    at conn_log (....\connection.js:707:23)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

then, the client do nothing until the game timeout.

wonderingabout:
i found interesting documentation about valid filenames in different operating systems:

https://stackoverflow.com/a/31976060
https://www.mtu.edu/umc/services/digital/writing/characters-avoid/

logfile fixed
logfile-fixed-2

tested to work on console and beta ogs
extra testing may be good to make sure we didnt make an accidental mistake

also unit testing for this will be added in #332 
fixes #337


wonderingabout:
this was implemented in #332 with unit testing and a few extra tweaks, closing this to avoid maintaining duplicates

wonderingabout:
this should be rebased on top of latest devel with latest updates and unit testing from #332 included

wonderingabout:
this was imported from #332 where i tested it on beta and wrote unit testing for it in config.test.js, so i assume it works, did not retest it one last time on beta though

utils/getValidFilename.js
```diff
@@ -0,0 +1,37 @@
...
+    const upperCaseCharacters = ["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z"];
+    const lowerCaseCharacters = upperCaseCharacters.map( (e) => e.toLowerCase() );
+
+    const allowedCharacters = singleFiguresToString.concat(upperCaseCharacters).concat(lowerCaseCharacters);
+
+    for (const char of validFilename) {
+        if (!allowedCharacters.includes(char)) {
+            validFilename = validFilename.split(char).join("-");
+        }
+    }
```
Dorus, line 32: This whole function is an awe full complicated way to write

```
return getUnreservedFilename(filename).replaceAll(/[^\w]/g, "-")
```

Also i dont like how this function is both calling an abstract `getUnreservedFilename` but also does a tons of string replaces. I would rather see
```
getValidFilename(filename) {
  const validFilename = getUnreservedFilename(filename);
  const charSafeFilename = clearInvalidChars(filename);
  return charSafeFilename;
}
```

config.js
```diff
@@ -247,6 +250,26 @@ function setRankedUnrankedOptionsDefaults(rankedUnrankedOptions, argv) {
     }
 }
 
+function exportValidLogfileFilename(argvLogfile) {
```
Dorus, line 253: `exportLogfileFilename` would be detailed enoughn i think

config.js
```diff
@@ -247,6 +250,26 @@ function setRankedUnrankedOptionsDefaults(rankedUnrankedOptions, argv) {
...
+        }
+        exports.logfile = validFilename;
+
+    } else {
+        // provide default filename based on start_time date and time
+        //
+        const filenameDate = `gtp2ogs-logfile-${exports.start_date.toISOString()}`;
+        const validFilenameDate = getValidFilename(filenameDate);
+        exports.logfile = validFilenameDate;
+        console.log(`Non-string logfile filename was successfully exported as ${validFilenameDate}`)
```
Dorus, line 269: Do we really need this else block and this string check?

Also i dont see how this would work if the argument is left out for filename. It will be an empty string then right? But still a string?

I think you mean to run the else block if no filename is supplied? if so, the code can still be refactored, first decided the filename and then put the duplicated instance of `exports.logfile = validFilename;` after the `if`.

utils/getValidFilename.js
wonderingabout: yes the abstraction was not so good, i wanted to make a 3rd function at first but i thought it was too much, but you're right it's better this way

also i would love to learn regexp, this is so awesome

except that i didnt find the replaceAll method, i read that if a regexp is supplied as input, the replace is global, but if a string is supplied as input in .replace then only the first occurence of the character is replaced, that's why i chose to use .split(char).join("-")
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

wonderingabout: i tested it and it works!

regexp replace

config.js
wonderingabout: sure :)

config.js
wonderingabout: oh indeed, logfile type should only be string, so we should add a .string in getArgv.js https://github.com/online-go/gtp2ogs/blob/devel/getArgv.js#L26

if only `--logfile` is used without anything, it would be a boolean and not an empty string

but then empty string would be tested false, so the code should slightly be adjusted to suppor that too

yes this can be improved a lot 🎉

utils/getValidFilename.js
wonderingabout: oh actually i tested more in details, and all works even than my code

regexp replace v2 works for all except underscore

the 1st source i linked especially mentionned whitelisted characters to include:

>Letters (a-z A-Z) - Unicode characters as well, if needed
>Digits (0-9)
>Underscore (_)
>Hyphen (-)
>Space
>Dot (.)

https://stackoverflow.com/a/29676339

wonderingabout: also yargs understands dots as objects and their properties, so better also replaced it with a hyphen too:
https://github.com/yargs/yargs/blob/master/docs/tricks.md#objects

wonderingabout: anyways after this PR is merged i'll learn regexp, wanted to do it for a long while now :)

wonderingabout: also actually about abstraction there is a nuance, reserved filename is an exact string match of a specific name (ex: COM1), so fixed reserved filename (ex: COM1-unreserved) does not need to be tested for invalid characters as ouput is known not to contain any

so it looks like 2 condtional branches actually, i'll refactor that :)

Dorus: But if you pump the strings trough both methods it'll be valid right? Why need to branch here?

Dorus: If anything i would remove the reserved filename thing, that seems like a user input error, the user should know what filenames his filesystem supports.

Dorus: Re: Yargs, the string has already been parsed by yargs at this point, i dont think we need to care about that anymore.

Dorus: Also yargs sees the flag as object, not the value.

Dorus: Another thing: Yargs can convert the input to number, but that is a valid filename, so the check for string is not correct.

Dorus: And since i'm writing a long list: Yargs seems to have really nice array support, we could add that for other options like komi's and boarsizes etc.

wonderingabout: yes yargs is very nice, we should definitely make use of some of its nice features like array and object support (also boolean --no-option flag for example)

wonderingabout: yes reseved filename is actually probably never going to be useful, and is system specific, so now the getValidFilename function is so small (2 lines) and also not used anywhere else, so i moved it it config.js instead of its own module

wonderingabout: so i added a .string in yargs's getArgv.js to make sure our input is string (number-only will be parsed as a string, and empty will not be a boolean true but should be an empty string (did not test it yet)

i'll import these changes in #332 and update unit testing for these

wonderingabout: also unrelated but yesterday i discovered git command-line in vs code is better than ubuntu shell command-line, with vs code's command line git remembers password and username, and most likely has other features too

vs code git command line

config.js
```diff
@@ -247,6 +249,30 @@ function setRankedUnrankedOptionsDefaults(rankedUnrankedOptions, argv) {
...
+    // to a hyphen (-)
+    return filename.replace(/[^\w]/g, "-");
+}
+
+function exportLogfileFilename(argvLogfile) {
+    if (argvLogfile === "") {
+        // provide default filename based on start date and time
+        //
+        const filenameDate = `gtp2ogs-logfile-${exports.start_date.toISOString()}`;
+        const validFilenameDate = getValidFilename(filenameDate);
```
Dorus, line 263: These 2 lines are duplicate, you can rewrite this as
```Js
function exportLogfileFilename(argvLogfile) {
        if (argvLogfile === "") {
            // provide default filename based on start date and time
            //
            argvLogfile = `gtp2ogs-logfile-${exports.start_date.toISOString()}`;
        }
        const validFilename = getValidFilename(argvLogfile);
        if (argvLogfile !== validFilename) {
            console.log (`Invalid logfile name "${argvLogfile}" has been automatically renamed`
                         + ` to "${validFilename}".\nValid logfile name can only be composed of`
                         + ` letters (A-Z a-z), numbers (0-9), hyphens (-), underscores (_).\n`);
        }
        exports.logfile = validFilename;
    }
}
```

config.js
```diff
@@ -247,6 +249,30 @@ function setRankedUnrankedOptionsDefaults(rankedUnrankedOptions, argv) {
     }
 }
 
+function getValidFilename(filename) {
+    // convert any other character than letters (A-Z a-z), numbers (0-9), hypens (-), underscore (_)
+    // to a hyphen (-)
```
Dorus, line 254: Actually regex \w means

> Matches any word character (alphanumeric & underscore). Only matches low-ascii characters (no accented or non-roman characters). Equivalent to [A-Za-z0-9_]

To also match hypen you would need `[^\w\-]`. (however since we convert to hypen this wont change the output.
Also we might want to allow dot and space, adding up to `[^\w\-. ]`

config.js
wonderingabout: yes, except that we modify our parameter here, would it be better to create a new const for that?

also, console.log message should be slightly adjusted (just saying that logfile has been renamed, no need to say it's invalid)
but yes this is better like this 🎉

config.js
wonderingabout: ah i see!!! i definitely want to learn regexp now 😮 

i think allowing space would be a bad idea (to begin with how do we add a space in the command line args), but the dots why not, even though just for a text filename that may be uneeded to have dots in name

Dorus: dot and space re valid filename charchers. What's bad about them?

wonderingabout: i thought space would be harder to use correctly in shell command line, and dot may change file extension in an unwanted way, but if these are not issues i can add space and dot too

config.js
wonderingabout: i made the no existing name check to a function is it good this way? (we reuse getValidFilename twice, is it clear that getLogfileFilename gets the initial logfile filename ?)

tested on beta to work, 

fixed logfilename

and also updated #332 and #332 's unit testing to match these changes
i find this code a lot cleaner than what i was :)

wonderingabout:
if we add testing for console.js it may be a good opportunity to also check a config file is indeed created at specified path (possibly delete it after test or use mocked fs that can handle that)

Dorus:
this one can be merged

PR #358, opened 2020-07-13
add unit testing for getArgv.js

wonderingabout:
allowed me to spot an extra space in rejectnewmsg default message in getArgv.js (already fixed in #332)
this also implements the fix in #357 

current code is quite ugly but this successfully tests yargs and default importing from process to getArgv.js

also i noticed after doing JSON.stringify in real gtp2ogs that empty object at index 2 should not exist so i sliced all process.argv elements starting from index 2, is it fine to fix it like this?

also test is surprisingly long (+/- 250ms), is it just the normal time it takes?

getArgv test result


anoek:
Thanks!

PR #359, opened 2020-07-14
move check_rejectnew to a local function in connection.js, add unit testing for rejectnew, rejectnewfile, rejectnewmsg

wonderingabout:
removed this chunk from #332 

tested rejectnew and rejectnewfile (with absolute and relative path) to work on beta (didnt test rejectnewmsg but unit testing is implemented in checkChallenge.test.js)

extra testing for this was added in #332 in config.test.js

should be ready to merge after we review this

### DIFF
--- a/connection.js
+++ b/connection.js
@@ -382,9 +382,9 @@ class Connection {
     }
     // Check bot is available, else don't mislead user
     //
-    checkChallengeBot(notification) {
+    checkChallengeBot(notification, fs) {
 
-        if (check_rejectnew()) {
+        if (check_rejectnew(fs)) {
             conn_log("Not accepting new games (rejectnew).");
             return { reject: true, msg: config.rejectnewmsg };
         }
@@ -519,18 +519,23 @@ class Connection {
         return { reject: false };  // Ok !
 
     }
+
     // Check challenge entirely, and return reject status + optional error msg.
     //
     checkChallenge(notification) {
 
-        for (const test of [this.checkChallengeSanityChecks,
-                           this.checkChallengeUser,
-                           this.checkChallengeBot,
-                           this.checkChallengeBooleans,
-                           this.checkChallengeAllowedGroup,
-                           this.checkChallengeHandicap,
-                           this.checkChallengeTimeSettings]) {
-            const result = test.bind(this)(notification);
+        const tests = [
+            [this.checkChallengeSanityChecks, notification],
+            [this.checkChallengeUser, notification],
+            [this.checkChallengeBot, notification, fs],
+            [this.checkChallengeBooleans, notification],
+            [this.checkChallengeAllowedGroup, notification],
+            [this.checkChallengeHandicap, notification],
+            [this.checkChallengeTimeSettings, notification]
+        ];
+
+        for (const [test, ...params] of tests) {
+            const result = test.bind(this)(...params);
             if (result.reject) return result;
         }
 
@@ -797,7 +802,7 @@ function processCheckedTimeSettingsKeysRejectResult(timecontrol, keys, notif) {
     }
 }
 
-function check_rejectnew() {
+function check_rejectnew(fs) {
     if (config.rejectnew)  return true;
     if (config.rejectnewfile && fs.existsSync(config.rejectnewfile))  return true;
     return false;
--- a/connection.js
+++ b/connection.js
@@ -1,5 +1,6 @@
 // vim: tw=120 softtabstop=4 shiftwidth=4
 
+const fs = require('fs');
 const http = require('http');
 const https = require('https');
 const querystring = require('querystring');
@@ -383,7 +384,7 @@ class Connection {
     //
     checkChallengeBot(notification) {
 
-        if (config.check_rejectnew()) {
+        if (check_rejectnew()) {
             conn_log("Not accepting new games (rejectnew).");
             return { reject: true, msg: config.rejectnewmsg };
         }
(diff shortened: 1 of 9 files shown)


test/checkChallenge.test.js
```diff
@@ -286,6 +286,111 @@ describe('Challenges', () => {
...
+        const notification = base_challenge({ ranked: false });
+
+        config.rejectnewmsg = 'Sorry, i am not available now.';
+        
+        const result = conn.checkChallengeBot(notification);
+        
+        assert.deepEqual(result, ({ reject: false }));
+    });
+
+    it('reject all games if rejectnewfile is used and a reject file exists', () => {
```
Dorus, line 355: Shouldn't we mock fs?

Dorus: Actually i'm not too sure, as long as we can cleanly refer to the right file and have no side-effect i think it's kinda ok to test like this.

wonderingabout: oh interesting i didnt think at all to write testing for `fs`

i read this and it seems interesting:
https://github.com/tschaub/mock-fs

i'll try to rewrite a mocked-fs version of the test

test/checkChallenge.test.js
```diff
@@ -286,6 +287,111 @@ describe('Challenges', () => {
...
 
+  describe('Rejectnew', () => {
+
+    it('reject all games if rejectnew is used with default rejectnewmsg', () => {
+
+      const notification = base_challenge({ ranked: false });
+
+      config.rejectnew = true;
+
+      const result = conn.checkChallengeBot(notification, fs);
```
wonderingabout, line 298: fs is now an external parameter that is passed to checkChallengeBot, so in the future if we want to write a fake fs we can easily do it in checkChallenge.test.js

should i add mock-fs (https://www.npmjs.com/package/mock-fs) to mock fs too, or is it fine to keep default fs in checkChallenge.test.js for now?

the only other occurence of fs in all our code is in console.js, so when we write testing for console.js, it may be a good time to make fs an external parameter too

connection.js
```diff
@@ -518,18 +519,23 @@ class Connection {
         return { reject: false };  // Ok !
 
     }
+
     // Check challenge entirely, and return reject status + optional error msg.
     //
     checkChallenge(notification) {
```
wonderingabout, line 526: i also refactored this part a bit to be able to provide different parameters for each function independently

i tested this to work on beta successfully too

test/checkChallenge.test.js
wonderingabout: also i tried to write manually a FakeFs class similar to FakeSocket but i was unable to do it :sweat_smile:

wonderingabout: oh actually it seems mock-fs is not used to simulate a filesystem's paths architecture, but it does not provide fs methods such as fs.existsSync

i read on the fs doc (https://nodejs.org/api/fs.html#fs_file_descriptors) that fs abstracts filepath to be operating system aspecific, so if we find a way to write a FakeFs that has our own fake method existsSync, would it work on all operating systems (ex: windows)?

Dorus: You can always use a sinon spy. However since right now we're only reading the file there is no direct need to do so yet. In fact using the real filesystem is acceptable for testing. These tests are usually called integration tests or end to end tests.

When you do have side-effects, you will need to make sure the test's initalise and clean up, do not collide etc. etc. Also we need to consider performance for the tests. For performance reasons and also test stability (test become very fragile when they all try to access the same resource and all need to initialise/clean up), it's usual to write unit-tests and those unit-test will instead use a fake/mocked instance of the io provider. This has the advantage you use a local instance and have no side-effects.

For the current tests we only read the files so it's actually ok to keep it as is. When we have a test that writes to the log, we must switch to a fake instead.

wonderingabout: ah i see, i get the main idea

so if we want to write testing for console.js, then we'd need to write a fake fs too at that time

so until then, is this pr good to merge, or should we take the opportunity to also write a fake fs (i'm interested how that can be done if we can do it in this pr)

Dorus: One other thing: Because of the way node imports work, it might be possible to replace the result of `require` calls in our tests, removing the need to insert includes like this in the constructor.

wonderingabout: ah but i didnt insert fs in the constructor, i tried to insert it in the constructor at first but it was uneeded to pass fs for all of our code if we only use it once in check_rejectnew

so i refactored this part of the code so that each checkChallenge subfonction can have its own parameters, not just notification:
https://github.com/online-go/gtp2ogs/pull/359/files#diff-0d74c7455b8cbe5be6f644690ecb1b8aR530
https://github.com/online-go/gtp2ogs/pull/359/files#diff-0d74c7455b8cbe5be6f644690ecb1b8aR385
https://github.com/online-go/gtp2ogs/pull/359/files#diff-0d74c7455b8cbe5be6f644690ecb1b8aR805


thats why in our tests we can pass fs only to checkChallengeBot, but we dont need to pass it in the connection constructor, nor in any other checkChallenge subfonction:
https://github.com/online-go/gtp2ogs/pull/359/files#diff-877056d4e4a6a8bc8dca4776a268c187R298

so gtp2ogs uses native fs, but if we want in our test we can pass a fake fs too now

i find it a lot nicer this way, what do you think

wonderingabout: i would still like to see how to fake fs (with sinon or some other way) if we can before merging this pr :)

wonderingabout: but since challenge always checks notification, we could slightly optimize it by writing notification only once and then the rest of params is everything other than notification

Dorus:
This one is ok to merge.

PR #360, opened 2020-07-14
minor cleanup, minor tweaks and fixes

wonderingabout:
minor fixes:
- remove fakerank and mingamesplayed from getArgv.js
- fix missing allow_all for speeds and timecontrols in config.js, fixed in a refactored way by adding exportInitialConfigRankedUnranked

other:
- changed indentation from 2 to 4 in gtp2ogs.js, if this is not fine i can revert it


wonderingabout:
tested on beta, currently there is an error while playing a game:

current error on beta

will investigate later

wonderingabout:
this also happens in devel, so one of the last commits introduced this

by hard resetting one by one until the issue disappears it can be found which commit created this issue

wonderingabout:
i found that after undoing this commit https://github.com/online-go/gtp2ogs/commit/c1bffb44ac308d83840b23883101fdb11b60eec5 the undefined error disappears, but bot still resigns after 5 crashes

before more recent commits should be unrelated to the issue

wonderingabout:
ok there is no errror looooooooooool

just my command line being invalid :sweat_smile: 

`cd ~/gtp2ogs_logs && node ~/gtp2ogs/gtp2ogs.js --beta --apikey woof --username meta--金毛测试-20b --persist --noclock --debug --greetingbotcommand --komis 7.5,automatic ../gtp2ogs/config.js -- ~/sai/build/sai-0.17-d2c82fc0 --gtp -w ~/networks/sai/9b/e1eab1d6_1913000.gz --noponder -v 400 --symm -r -1 --lambda 1.0 --mu 0`

`../gtp2ogs/config.js` here rejectnewfile was missing: `--rejectnewfile ../gtp2ogs/config.js`

wonderingabout:
tested to work on beta, a game works end to end :)
awaiting review before we can merge this

config.js
```diff
@@ -79,30 +57,17 @@ exports.updateFromArgv = function(argv) {
...
         exports.logfile = `gtp2ogs_logfile_${new Date().toISOString()}`;
     }
-    for (const k of ["timeout", "startupbuffer"]) {
-        if (argv[k]) {
-            // Convert some times to microseconds once here so
-            // we don't need to do it each time it is used later.
-            exports[k] = argv[k] * 1000;
-        }
-    }
+    exportMicroseconds("timeout", argv);
```
Dorus, line 60: `exportMicrosecondsAsSeconds` is a more accurate name.

config.js
```diff
@@ -79,30 +57,17 @@ exports.updateFromArgv = function(argv) {
...
-    if (argv.minhandicapranked > -1) {
-        exports.noautohandicapranked = true;
-    }
-    if (argv.minhandicapunranked > -1) {
-        exports.noautohandicapunranked = true;
-    }
-
+    exportNoAutoHandicapIfMinHandicapIsPositive("", argv);
+    exportNoAutoHandicapIfMinHandicapIsPositive("ranked", argv);
+    exportNoAutoHandicapIfMinHandicapIsPositive("unranked", argv);
```
Dorus, line 69: Wrap these 3 in a function.

config.js
```diff
@@ -151,23 +116,35 @@ exports.updateFromArgv = function(argv) {
     // console messages
     // C - test exports warnings
 
-    testExportsWarnings();
+    warnIfNopauseIsMissing("ranked");
+    warnIfNopauseIsMissing("unranked");
```
Dorus, line 120: No need for general arg? Also, these can be called in a single function.

config.js
```diff
@@ -151,23 +116,35 @@ exports.updateFromArgv = function(argv) {
...
 function testRankedUnrankedOptions(rankedUnrankedOptions, argv) {
     for (const option of rankedUnrankedOptions) {
         const [general, ranked, unranked] = getArgNamesGRU(option.name);
         
         // check undefined specifically to handle valid values such as 0 or null which are tested false
         if (argv[general] !== undefined) {
             if (argv[ranked] !== undefined) {
-                throw `Cannot use --${general} and --${ranked} at the same time.\nFor ranked games, `
-                      + `use either --${general} or --${ranked} or no option if you want to allow all values.`;
+                throw `Cannot use --${general} and --${ranked} at the same time.\nFor ranked games, use either --${general} or --${ranked} or no option if you want to allow all values.`;
```
Dorus, line 144: No option means we use default right?

config.js
Dorus: Nevermind, it checks genral arg twice, in both calls. Mmm.

config.js
wonderingabout: oh actually it seems we were all mistaken
https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md#timeout

this exports time in seconds input as milliseconds output
so i guess `exportSecondsAsMilliseconds` is the correct name?

config.js
wonderingabout: fixed :100:

config.js
wonderingabout: should i also wrap this in a function?

config.js
wonderingabout: the `or no option` in `or no option if you want to allow all values` was not clear indeed, plus it does not always allow all values, it allows all values only if the option's default is empty

i replaced it with `or do not use anything to switch to default settings for this option`
it is indeed a lot clearer :)

config.js
```diff
@@ -253,15 +230,30 @@ function getValidFilename(filename) {
...
+function exportLogfileFilename(argvLogfile, argvDebug) {
     const filename = getLogfileFilename(argvLogfile);
     const validFilename = getValidFilename(filename);
     
-    if (filename !== validFilename) {
-        console.log (`Logfile name "${filename}" has been automatically renamed to`
-                        + ` "${validFilename}".\nValid logfile name can only be composed of`
-                        + ` letters (A-Z a-z), numbers (0-9), hyphens (-), underscores (_)`
-                        + `, spaces ( ), dots (.).\n`);
+    if (argvDebug && filename !== validFilename) {
```
wonderingabout, line 255: i also made the logging message optional only if debug is used

config.js
Dorus: Did you?

config.js
Dorus: yeah that would be nice.

config.js
```diff
@@ -144,8 +105,30 @@ exports.updateFromArgv = function(argv) {
...
+
+function exportInitialConfig() {
+    exportInitialConfigRankedUnranked("");
+    exportInitialConfigRankedUnranked("_ranked");
+    exportInitialConfigRankedUnranked("_unranked");
 
+    exports.start_date = new Date();
+}
+
+function exportInitialConfigRankedUnranked(rankedUnrankedUnderscored) {
```
Dorus, line 122: I think the name `exportInitialConfig` would be good enoug.

config.js
wonderingabout: oh i did it only for exportInitialConfig() at line 13
sure i can do it here too 🛰️

config.js
wonderingabout: roger 🎖️

config.js
wonderingabout: but we already use this name at line 114 for the wrapper function, also exportInitialConfig of line 114 does other things than just the _ranked exports, it also exports start date for example, and there we could add other initial config exports in the future

so maybe exportInitialConfigUnderscored for the sub functions?

config.js
and it does look a lot nicer :100: :open_mouth: :100: :trophy:

config.js
wonderingabout: fixed

config.js
wonderingabout: fixed, it is indeed clearer ranked unranked could be any option, but this function is only for the underscored (`_ranked`, `_unranked`, ``) initial exports

wonderingabout:
pulled latest devel and tested this to work on beta
should be good to merge :1st_place_medal:

config.js
Dorus: Ah yes i missed the other function. But i like the new name, nicely found :D

config.js
```diff
@@ -113,39 +72,37 @@ exports.updateFromArgv = function(argv) {
 
     // 2) specific ranked/unranked options exports
 
-    processRankExport("minrank", argv);
-    processRankExport("minrankranked", argv);
-    processRankExport("minrankunranked", argv);
+    exportMinMaxRanks(argv);
+    exportBansUnderscored(argv);
+    exportAllowedGroupsUnderscored(argv);
```
Dorus, line 77: Why the need to mention underscored on these 2 funcitons?

config.js
```diff
@@ -359,20 +347,27 @@ function processAllowedGroupExport(argName, argv) {
...
-        }
-    }
+function exportAllowedGroupsUnderscored(argv) {
+    processBoardsizesExport("boardsizes", argv);
+    processBoardsizesExport("boardsizesranked", argv);
+    processBoardsizesExport("boardsizesunranked", argv);
+
+    processKomisExport("komis", argv);
+    processKomisExport("komisranked", argv);
+    processKomisExport("komisunranked", argv);
```
Dorus, line 357: We probably want to change this in a speparate pr, but...

It's strange to call a function called `processKomisExport` with `("komis` every time. This could be simply

```
    processBoardsizesExport("", argv);
    processBoardsizesExport("ranked", argv);
    processBoardsizesExport("unranked", argv);

    processKomisExport("", argv);
    processKomisExport("ranked", argv);
    processKomisExport("unranked", argv);
```

config.js
wonderingabout: indeed this is redundant
i guess the first name exportInitialConfigUnderscored was really nice and i got a bit carried away :sweat_smile:

config.js
wonderingabout: yes, but this makes processRankExport and processAllowedGroupExport easier to read as they all share the same input parameter type (full optionName +/- rankedunranked)

it may be fine to keep it as it is, do i change it according to this comment?

Dorus: I'm not talking about `processRankExport` and `processAllowedGroupExport`, those functions are just fine. I'm talking about `processBoardsizesExport` and `processKomisExport`.

Dorus:
This can be merged.

config.js
wonderingabout: yes these should be changed, but as this pr is already crowded we may instead do it in another pr in the future as you suggested

PR #361, opened 2020-07-18
fix ranked unranked defaults incorrectly exported

wonderingabout:
unit testing for this is included in #332 
was tested to work on beta too in #332


Dorus:
This can be merged. @roy7 must like this one :D

PR #362, opened 2020-07-18
fix errors not thrown before exporting

wonderingabout:
unit testing tested this to work in #332 
was tested to work on beta in #332 

code is not cleanest and may need review though :sweat_smile:


wonderingabout:
can be adressed in a lot simpler way as explained in https://github.com/online-go/gtp2ogs/pull/332#discussion_r464050773
will be done directly in #332

PR #363, opened 2020-07-22
add --noprovisional --noprovisionalranked --noprovisionalunranked

wonderingabout:
reject message would need tweaking (a bit too long), did not test it yet on beta, but i added unit testing so it should work
also number of ranked games estimated needed was for maxrankdeviation, but do we also want to add it here?


Dorus:
Can be merged

PR #369, opened 2021-01-24
Improved time settings

Dorus:
This change the way we communicate time_settings to the bot.

It adds auto detect for support of kgs time, kata time (for fischer) and fixes a number of bugs on existing time settings.
It also adds unittesting for the time_setting code.

A non exhaustive list of bugfixes:
Under fischer time the bot believed to have the time_increment already added during the current move.
Under simple time, white always believed to have only 1 second.
While going from main time to period time, the reported main time could go negative. This happened most notable under japanese byoyomi, and could cause timeouts as most bots will give an error on negative time input. An easy way to trigger this bug before would be to turn off the bot during byo-yomi and restart it after at least 1 period expired.
In many cases `time_settings` was modified to communicate remaining time, instead of `time_left`.

### DIFF
--- a/bot.js
+++ b/bot.js
@@ -45,7 +45,7 @@ class Bot {
         }
 
         this.proc.stderr.pipe(split2()).on('data', (data) => {
-            if (this.ignore)  return;
+            if (this.ignore) return;
             const errline = data.toString().trim();
             if (errline === "") return;
             this.error(`stderr: ${errline}`);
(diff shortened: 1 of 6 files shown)


roy7:
@anoek Would you like to give the bot.js changes a once over to see if anything jumps out at you? Or give this PR a test spin on your bots for a bit before we merge? Dorus is running successfully on his bots so far.

anoek (approved):
This looks great, thanks for all of the hard work @Dorus!

PR #372, opened 2021-03-07
Added sanity check for move_number

Dorus:


roy7:
@anoek Dorus has seen an issue in a few games where the move packet is sent more than once from the server for the same move. Not sure if this is a new server bug or when it started. He'll test this a bit then we'll merge it as a workaround/failsafe on the gtp2ogs side. I assume server should never be sending a move packet more than once.

roy7:
@Dorus Could add a bit more verbosity to the error message if you want, like `expected ${this.state.moves.length} + 1`. So a quick glance at your log you see what it expected as a sanity check.

Dorus:
@roy7 Done. I was relying on the previous log line myself, but i guess you're right, better to include it for easy access.

Dorus:
Been a week and i haven't seen the log line from this bug in my gtp2ogs. Possibly @anoek did solve it?

Anyway, no log line does also mean this solution is safe, and we've seen the problem 3 times in under a week. So i'de say we should merge this just to be safe, and in case the cause of the bug is more rare but still exists. @roy7

anoek:
Cool, I did fix a race condition that might have been the cause of this a bit ago, and I think that landed probably about a week ago or so, so it might be fixed, but I'm all for sanity checks.

PR #390, opened 2023-01-09
Add support for a secondary "Resign Bot"

anoek:
This PR adds support for a new flag `--resignbot "<cmd> <args>"`, which when specified will start up a second bot command to consult for each move, if the resulting move is a resignation, the bot will resign.

The goal of this is to allow weak bots, who sometimes have trouble realizing they should resign, to use a second presumably much stronger bot to handle resigning in a timely manner.

This PR also contains some non functional refactoring to switch from callback to async style processing in much of our bot and game code, which made adding this feature notably easier and reduces the overall complexity of the code base.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,7 +13,6 @@ module.exports = {
         "@typescript-eslint",
         "@typescript-eslint/tslint",
         "prettier",
-        "react-hooks",
     ],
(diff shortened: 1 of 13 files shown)


anoek:
Closing in favor of v8 which incorporates support for this

PR #391, opened 2023-04-11
gtp2ogs V8

### LINKED ISSUE #28: Options for max and min rankedhandicap and 'auto' handicaps

### LINKED ISSUE #32: Is phase==play pass still needed?
I saw this happen in a game, I assume because of a server pause situation or a player pausing and un-pausing. It printed the 'lazy' message to the log file.

gtp2ogs at the moment doesn't look at pause/play states when deciding if it should move. If a new connection/gamedata is our turn, we make a move. If a move event comes in that was the opponent's move, we make a move. The bots play fine during a paused game.

But if a human opponent moves, starting bot's makeMove/genmove process, and then unpauses, the phase==play event would trigger the bot to immediately pass. This would cause internal bot state with --persist to be wrong (it generated a move the server didn't get).

I'm just not certain on all situations a phase==play event might be sent from the server and if there's some special situation(s) that actually do need handled? Or can we just comment that out for now and not pass when games are un-paused?

It also seems if game scoring doesn't go through, and we move back to phase play from phase stone removal, we'd immediately pass again.

This is the section around line 661.

### LINKED ISSUE #37: Unknown command errors can interfere with callback buffer processing
Had a situation where someone used --kgstime with RayN, which doesn't know kgs-time_settings command. This caused an "? Unknown command" response from the bot. After this, other valid commands with proper response (such as genmove for first move of the game) did not get processed correctly.

I'm assuming the unknown command response didn't get consumed off the stdin buffer properly, since issuing the commands that resulted in error weren't expecting any response. Need to be sure any unknown command errors are either cleared and then the next step in buffer read by a later command, or some sort of clean error message is reported to the log.

### LINKED ISSUE #56: Add --pausemaxtime
Add an option to allow pausing but with a max time limit. Handle it like --nopause but SetTimeout() instead of resuming immediately. If game resumes, clear the timeout.

### LINKED ISSUE #85: Issues with slow startup of bot
I am new to OGS and gtp2ogs and tried to set up a bot for regular games. I use a command similar to the following for this purpose:

    pushd C:\path\nodejs && node.exe C:\path\nodejs\node_modules\gtp2ogs\gtp2ogs.js [...] \
            --startupbuffer 20 --persist -- bot bin args

This works when the game has sufficient time. But when playing blitz, the bot takes too long to start and times out before making the first move.

Apparently the bot process is only started once the game begun. Is there a way to start it immediately, before connecting to OGS? Then it could react in an instant without booting up.

### LINKED ISSUE #100: 404 error Challenge does not exist
So in the last days i noticed the "challenge" does not exist error happens a lot !!

see my recent messages and screenshots here : 
https://github.com/online-go/gtp2ogs/issues/52

this week, i got this error 4 times, thats a lot lol (didnt screenshot them all)

below is a summary of these errors

day 1 (18 january)

challenge does not exist loop
challenge loop2

for this first time the challenge looped (hundreds of errors for hours)
scrolling ogs notifications, i noticed one challenge was not deleted automatically by the bot (main time too short)

then the other times, i checked but there was no challenge to be deleted on the notifications (but still triggered the "challenge does not exist error" once)

day 2 (19 january) 

challenge error3

day 3 (yesterday) : 

challenge error

day 4 (27 january) 

challenge error4

@roy7 @tokumotoK @Dorus @windo @lemonsqueeze 

any idea why and how to fix ?

this can be very confusing for new gtp2ogs users

### LINKED ISSUE #107: Startupbuffer better replacement
This was bigger than I wanted to try and tackle so I've ignored it, but the --startupbuffer I added as hack to handle the fact the time_left command is sent to the bot while it's starting up on that first move, but the bot won't read the command until the initial startup has completed. This means by the time the bot reads the time_left command, some time has been spent and the info is outdated.

A way around this might be to detect when the prior commands like 'komi' have returned the empty '=' response (so we know bot has initialized and is awake) and then send time_left using actual current time left. In the Chess world the engines seem to have an is_ready command they send and the clock doesn't start until the bot says it is ready.

LZ might eventually create multiple startup threads to process GTP commands even while initializing but a more flexible solution for all bots would be best for gtp2ogs to try and support. Perhaps just sending a mandatory GTP required command (list_commands?) at bot startup and not send the time_left/etc until we have that response is a solution.

Posting this as a reminder one day it maybe needs improved.

### LINKED ISSUE #112: Cleanly support a wide range of time settings.
Admin allows time systems: none, absolute, simple, fisher, canadian and/or byo-yomi.
For none, no future checks are required.
For absoluate, only min and max main time needs to be checked.
For simple, only min and max period time needs to be checked.
For fisher, you need to check initial time against min and max main time. You need to check added time vs min and max period time.
For canadian, you need to min and max main time. You need to check time per x stones against by taking (period time divided by x) and compare against min and max period time. You need to check x stones also against min/max nr of periods.
For byo-yomi you need to check min and max main time, min and max period time, and min and max periods.

So bot admins only need to set allowed `time systems`, `min` and `max main time`, `min` and `max period time`, `min` and `max periods`. A total of 7 settings to support everything.

### LINKED ISSUE #125: Config file containing all arguments instead of command line
so

as @windo suggested in https://github.com/online-go/gtp2ogs/issues/119#issuecomment-461848202 , i thought it would be a very nice idea to use a config file instead of a bunch of randomly lineared argv

what motivated me to do that is to add `--minmaintime` `--minperiodtime` differently for blitz,live, corr games etc...
for example `maxmaintimelive 600 maxmaintimecorr 86400`
so its about time we do this config file i think

easily readable, easily exportable, easily expandable to add new features to config, etc..

i dont have any knowledge on the topic, but i'm willing to try and learn,

this is the documentation @Dorus sent me on discord : 
- https://github.com/lorenwest/node-config/wiki/Common-Usage#using-config-values
- https://github.com/lorenwest/node-config/wiki/Configuration-Files
- https://www.npmjs.com/package/config

will take some time for me to do it, maybe 1 or 2 weeks to release a pre-PR, but i think it would definitely be nice if doable before next master release, see #119 

@roy7

### LINKED ISSUE #136: Auto hiding bots at connection cap
Just a feature idea, if a bot has a max connection limit and has hit that limit, should it not appear in the drop down list any more? A complexity is it might be capped out for live games but still allow corr games and we can't control that on the server side, we either are hidden or we aren't. Perhaps if we have a connection limit and don't allow corr, then auto hidden makes sense.

### LINKED ISSUE #139: server disconnects can cause games to get stuck

### LINKED ISSUE #200: reminder to fix (and maybe add) some text messages (sendchat) with the human player
in our attempt to make gtp2ogs more talkative with the end user using more sendchats (which i think is great), there are some text messages that dont work as intended

this one works fine for example : https://github.com/online-go/gtp2ogs/blob/devel/game.js#L209

but this one for example pops at the wrong time and multiple times : 
https://github.com/online-go/gtp2ogs/blob/devel/game.js#L269
https://github.com/online-go/gtp2ogs/blob/devel/game.js#L319

for example : https://beta.online-go.com/game/6300

it is easy to produce these with phoenixgo : "just start a game with any komi different from 7.5 (for example 6.5) and the engine will crash" :100: 

2 problems : 
1) ORDER : the messages pop in the wrong order (resigning then trying to restart then resigning)
2) TIMING : the messages dont pop until the game is finished : we'd want the "trying to restart" messages to pop at the first crash, then the 2nd, etc. until the (n-1)th crash, AND then at the (n)th crash we'd want only one resigning message instead

i added these sendchats in a rush and forgot to fix them, so this is a reminder for that

finally, besides that, we may want to add a few extra sendchats if it helps the human player (moving greeting earlier was a nice idea too for example), but of course we wouldnt want to spam
(we may add also add a `--talkative` option or a `--silent` gtp2ogs option depending on which we want to be default)

### LINKED ISSUE #276: Refactor loadClock
in bot.js, loadClock has a lot of repetitive code

it is a bit sensitive to bugs code, so we dont want to change it too drastically until we give it extended testing and review

this issue is also a reminder that we made an attempt of refactoring in #223 that may serve as a reference or comparison in the future

### LINKED ISSUE #278: Show all reject reasons
When gtps2ogs reject a game, it only shows the first reject reason.

It would be better to show all reject reasons.

### LINKED ISSUE #279: replace ranked / unranked repetive patters with r_u, r_u_strings, config_r_u
in config.js and connection.js, there are a lot of repetitive patterns:

- generalArg
- rankedArg
- unrankedArg

instead of checking the 3 separately, we could use the `r_u` approach (for ranked_unranked)

- if game is ranked check general else ranked arg, and display not allowed for ranked games
- if game is unranked check general else unranked arg, and display not allowed for unranked games

user doesnt need to know the detailed setup of botadmin, only if he can play

the `r_u` code can be looked into in #230 

it also includes r_u_strings that simplify reject messages by generation only once all ranked specific "for ranked games" (for r_u games)

there's a further advanced version of this idea, which is to export everything in the format
`config_r_u` which is:

- config.ranked
- config.unranked

then load config_r_u at challenge check and greatly simplify arg checks in connection.js, for example if (config_r_u.maxhandicap which can be either config.ranked.handicap or config.unranked.handicap based on notification.ranked)

edit: this is currently done in #320

### LINKED ISSUE #280: add --whitelist
we currently have --bans (which is a blacklist of users and user ids), but some bot admins expressed to wish to also have a whitelist

it could serve various purposes and is fine to add

an implementation was attempted in #232 

we can work on it again and only pr the changes specific to that code without adding other changes

### LINKED ISSUE #282: Add increment and range operators for numbers (ex: 9:13 for 9,10,11,12,13)
this was done in #230 

we'd need to do it in a specific pr, after we remove all the repetitions for general/ranked/unranked exports of allowed families in config.js

### LINKED ISSUE #284: close and open bots at specific times (ex: at 09:30 PM)
it would be nice to be able to close and open our bot with a schedule, so that we dont have to worry of accepting games before sleep time / shutdown time / scheduled leave

then automatically reopen the bot with a timetable for everyday of the week
an attempt was made, almost finished but not tested in #238 

we may look at it as a future reference or comparison

### LINKED ISSUE #288: TODO: redefine time defaults to actually match blitz live corr edge cases
currently, when user sends a challenge on OGS, it is sometimes of a different speed type, for example:

- blitz byoyomi maintime 5 minutes is sent as a live challenge
- live byoyomi periodtime 1 hour is sent as a correspondence challenge

this is because the boxes in OGS website allow to input values that are out of range of the blitz live correspondence edge values

to find the accurate edge values, we can do something like --maxmaintimeblitz 1 --maxmaintimelive 999999999 and gradually increase maintime until the challenge is not rejected anymore

this will be the edge value when a blitz challenge is converted to a live challenge

we can also send an issue to @anoek on OGS's github and hope the boxes are fixed to match actualy blitz live corr values

### LINKED ISSUE #307: fix console log methods in .js files
@Dorus explained on the discord:

> gtp2ogs doesn't even use console, it uses tracer that uses console.
https://www.npmjs.com/package/tracer
in a few cases our classes have extra log methods

for example in logfile, @windo uses console:
https://github.com/online-go/gtp2ogs/blob/devel/console.js#L27-L37

>yes thats cool.
He takes the actual console, and hooks into it to write to the file
But it's not used to log things
it's used to read what's logged.
The writing to it goes trough tracer package
however in a few cases our classes have extra log methods
So it will always give that line number
it would be better if it gave the calling line number
I think we should also fix that sometimes. Either remove the log method, or make it report the line number one higher in the stack trace
Ah i think it's the stackIndex property
Would need to experiment with that

adding this as an issue to keep this great idea in mind!! :100:

### LINKED ISSUE #311: check username and apikey are registered on OGS before displaying a "Connected"
currently if i use `--username woof --apikey woofwoof` we still connect successfully

this username and apikey do not exist on ogs though, so in case bot admin enters  typoed username/ apikey, or just in case of misuse, it would be good to add a check for that and throw an error instead of connecting

### LINKED ISSUE #321: allow "all" is redundant, no allow "all" force sets defaults
while refactoring #320 , i noticed these 2 current problems

they can't easily be changed and they show we're hitting the limit of command line argv, so it seems we'd really want to work on a config file after #320 

now that this was said, here are the 2 observations i noticed:

## 1)

we now (and in the past too) export an allowed_boardsizes, allow_all_boardsizes, argv (config.boardsizes)
now only the format changed  but it's still the same principle:

```Js
             boardsizes:      { allow_all: false, allowed: [], argv: undefined },
```

but actually the "all" value is useless, if we want to allow all boardsizes or all komis or all timecontrols, we should just not set any arg instead of bothering with an extra property i think
so we may just export

```Js
boardsizes:      { allow_all: false, allowed: [], argv: undefined }
```

or even better
```Js
boardsizes: []
```

we currently use config.boardsizes for parsing allowed boardsizes ("9,13,19" -> "allowed board sizes are 9x9, 13x13, 19x19"), and same for speeds and komis etc.
but actually we dont really need the string argv itself, we have all the information in 
`Allowed speeds are ${Object.keys(config.speeds)}.`;

if user wants to allow all speeds, he can just simply not use --speeds!
the --boardsizes all is an artifact from the past when windo added komis support for all values, and then i used it to have "all" or "custom" or "9,13,19"
this is the code i initially added for komi support:
https://github.com/online-go/gtp2ogs/commit/75823d38d222f1f7019ceb6d3b08a64c09f35924#diff-9b3d294c3963ab54d85d79ba5d9ba223R158-R162
then windo added komi any
https://github.com/online-go/gtp2ogs/commit/b945e2f4ddddddfe1a79b60316e748ed3e7bf09c#diff-9b3d294c3963ab54d85d79ba5d9ba223R241-R250
but now that i think about it, it really seems like we can just not use the arg if we allow all values, same as if we don't set --minrank or --maxhandicap we allow all values, we dont do --minrank all

## 2)

but there is a problem with this
for example boardsizes has a default 9,13,19
so if we dont set any arg and want to allow all values, the default system will forcefully set the default even if we want to allow all values
also, if we set only a ranked value --boardsizesranked 19 then the new defaults system in lastest commits will also set it to unranked, so botadmin will not be able to allow all boardsizes in unranked without also setting --boardsizesunranked all
but this seems super hacky and not convenient
if we add a default to a new option in the future, or actually even just looking at maintime and periodtime, if botadmin wants to allow any value in unranked and defaults in ranked it's literally impossible since we force set defaults in unranked if they are only in ranked, and there are no --minmaintimecorrunranked all option
so instead of adding an "all" option to all args, we may just remove it everywhere and set a system where we just dont use an arg if we want to allow everything
to do that, we need to rethink how we want our defaults to work and be set
especially on commandline, it would make command much lighter by not setting "all" for every option we dont care about
i think we're reaching the limit of how far command line can get us 
it seems we really need a config file lol
so for now i'll leave it as it is

but hopefully after we merge my current pr, if all is well, our next task should really be config file!

### LINKED ISSUE #347: make allowed boardsizes and allowed komis objects, not arrays
currently allowed komis and allowed boardsizes are arrays of allowed indexes, but it would be easier to write them as objects of allowed keys that have the value true, same as speeds and timecontrols

this is currently being done in #320

### LINKED ISSUE #383: [FeatReq] add an option to use a config file instead of writing all args in CLI
## Context
As current options can only be inputed as CLI args and the number of options seems to grow,

> In gtp2ogs, options need to be inputted as command-line arguments.
> The number of options (and of bot arguments) is free, use the ones you want.
Origin: https://github.com/online-go/gtp2ogs/blob/devel/docs/OPTIONS-LIST.md

Adding a file to store it can be convenient for

1. fist-time setup
2. TS typing
3. profile different default config (like a `katago_config.ts` for katago users)
4. switching between configs
5. handling space character (` `) in username

## Suggestion

We can use either dot-env, a json file, or a TS/JS file to store all the configs (like [ESLint Config](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new)). I'd prefer to first support TS file to give user a clue without adding JSON-Schema.

The file would be like
```ts
interface OGS-Config {
beta: boolean;
host: string;
port: number;
apiKey: string;
maxconnectedgames?: number;
debug?: boolean;
// ...
}

export default const config = {
host: "online-go.com",
port: 443,
beta: false,
apiKey: "XXX_SOME_API_KEY_XXX"
}
```

To adopt this config we need to

1. add a `--configFile path/to/config.ts` or use a default path.
2.  update the function `updateFromArgv`
    https://github.com/online-go/gtp2ogs/blob/d275c684d68630a1311fca09a6ca8a7f4f664bad/src/config.ts#L14

## Side-effect

1. more readability for `updateFromArgv`
2. easier type-checking and more ts adoption
3. more structural config
6. we can have a constant `DEFAULT/CONFIG`/`BASE_CONFIG` and split the default/suggested config from building a customised config. 

## Alternative

Create an executable `run.sh` in repo root like below (sample args from [tutorial-3A5](https://github.com/wonderingabout/gtp2ogs-tutorial/blob/master/docs/3A5-linux-run-gtp2ogs-js-beta.md)

```sh
node ~/gtp2ogs-node/node_modules/gtp2ogs/gtp2ogs.js \
--beta \
--apikey 53e4288597ab04c591dffikz4se8u6v1y2z8tty7 \
--username meta--金毛测试-20b \
--debug \
--persist \
--noclock \
-- \
/home/amd2020/PhoenixGo/bazel-bin/mcts/mcts_main \
--gtp \
--config_path=/home/amd2020/PhoenixGo/etc/mcts_1gpu.conf
```

anoek:
This is a complete overhaul of the gtp2ogs system to better support bots going forward.

 * All configuration is now done through config files
 * Acceptable game settings is now sent to online-go.com as part of the authentication process so that the user interface can guide users through acceptable parameters
 * Add support for both opening and ending bots to better support weak bots
 * PV output is now sent in a way that will enable the UI to render the PV output as a graph instead of a stream of text in the game chat (though that is still the behavior of the UI as we transition)
 * Translated greeting, farewell, and rejection notices
 * Inform the server when we are at capacity so the UI can act accordingly
 * Automatically unpause games after a certain time
 * Add bot process pooling instead of one process per game
 * Switch to Goban direct websocket with type checked client to server messages 
 
 

Fixes #383
Closes #347 
Closes #321 
Closes #311 
Fixes #307 
Fixes #288 
Closes #284
Fixes #282
Fixes #280 
Closes #279 
Fixes #278
Closes #276 
Closes #200 
Fixes #139 
Fixes #136 
Fixes #125 
Fixes #112 
Fixes #107 
Closes #100 
Fixes #85 
Fixes #56
Closes #37
Closes #32 
Fixes #28


roy7:
Wow you've been busy. :)

PR #412, opened 2023-09-28
Add container dev environment, pkg instructions

raylu:


raylu:
would still like to get this merged

roy7:
We can always change it more later if something here wasn't quite right. Better than leaving it stale. :) Thanks for the work @raylu

Issue #258, opened 2020-05-06
collect user feedback from ingame chat (--loguserchat)


wonderingabout:
it would be helpful to auto collect user feedback from all game chats

most humans dont message bot admins or bot accounts privately, and bot admin may miss these messages

so we may want instead a way to fetch user messages from ingame chat, similarly to what was done in #216 
 
we may add an option to collect user game chat and gather it in a file
most people are silent against bots so we dont risk overflowing at all (and can add safeties against that)

instead of (message my admin or user X in case of errors), we could just do:
```
please write any feedback or request regarding this bot here, the bot admin will see it and may reply to you
```

then we'd just need to find a way to filter out the pv and other bot-customized sendchats like kata-bot or the katrain bots have

it would be an option (not default) `--loguserchat` (to file)

(we may even add as an argument maximum allowed text length, for example --loguserchat 100, or we could hard code this max length limit but it would be less interesting i think)

anyways in all cases it would exclude all total game message strings that are longer than 100 words
to auto filter pv, aichat, bot sendchat, and human spam

what do you think?

Dorus:
I still like this idea.
Similar: Currently the bot auto-removes all notifications it doesn't understand (and i think also all those it does understand?) An option to ignore the "late chat received" notification instead of dismissing it, could be useful.

windo:
Ooh, I also like this idea a lot. I haven't touched gtp2ogs in a while, but it's on the top of my list of things to do if I should ever want to hack on it again.

Dorus:
Related to this: I *might* have accidentally removed all 'old chat notifications' on bot accounts when i added logic that removed all ogs unknown messages on bot accounts.

Ah first quick step on this issue might be to make chat notifications known again and not discarded, or make this an option for bot owners that want to read chat msgs in games.

windo:
I've been running this for a week and I see some messages duplicated. I guess I made some mistake with (or didn't understand something about) the message timestamps.

windo:
And as for the chats themselves, looks like they are mostly:

1. Variations of "hello" or "thanks for the game" (though ofc most people don't greet the bot)
2. Being annoyed that the game is not going very well
3. Asking for an undo
4. Being surprised by the scoring

I think I'll plan to review the chats ~weekly or so for the next little while.

raylu:
> I've been running this for a week and I see some messages duplicated. I guess I made some mistake with (or didn't understand something about) the message timestamps.

[...]

PR #421, opened 2024-05-15
Handle game chats and write them into the log

windo:
Players sometimes say things in the game chat, which I would be interested in at least becoming aware of.

This ought to address https://github.com/online-go/gtp2ogs/issues/258

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -26,6 +26,7 @@ export class Game extends EventEmitter<Events> {
     state: GoEngineConfig;
     opponent_evenodd: null | number;
     greeted: boolean;
+    startup_timestamp: number;
     bot?: Bot;
     using_opening_bot: boolean = false;
     ending_bot?: Bot;
@@ -56,6 +57,7 @@ export class Game extends EventEmitter<Events> {
         this.verbose = trace.debug.bind(null, `[game ${game_id}]`);
         this.warn = trace.warn.bind(null, `[game ${game_id}]`);
         this.error = trace.error.bind(null, `[game ${game_id}]`);
+        this.startup_timestamp = Date.now() / 1000;
         this.state = null;
         this.opponent_evenodd = null;
         this.greeted = false;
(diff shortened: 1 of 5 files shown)


windo:
I'm not entirely confident that it wouldn't be a better idea to still try to just `connectToGame` from main and make sure `Game` can see the game is finished and will disconnect?

Also, should this be optional?

windo:
Oof, I've messed something up and it doesn't pick up chats during the game, will check later.

windo:
Ought to be fixed. I'm running it on my bots now and they log when I chat with them during or after a game.

src/Game.ts
```diff
@@ -827,6 +848,20 @@ export class Game extends EventEmitter<Events> {
...
 }
 
+export function handleChatLine(game_id: string, line: any, cutoff_timestamp: number) {
+    if (typeof line.body !== "string") {
+        return;
+    }
+    if (line.username === config.username) {
+        return;
+    }
+    if (line.date < cutoff_timestamp) {
```
raylu, line 858: dunno if it's worth noting that converting `new Date()` to a number gives you epoch time in UTC and `line.date` is also epoch time in UTC

src/Game.ts
```diff
@@ -341,8 +344,26 @@ export class Game extends EventEmitter<Events> {
...
+                // game has already been dealt with.
+                handleChatLine(game_id, d.line, this.startup_timestamp);
+            };
+            socket.on(`game/${game_id}/chat`, on_chat);
+            this.on("disconnecting", () => {
+                socket.off(`game/${game_id}/chat`, on_chat);
+            });
+        }
+
+        trace.log("connecting");
```
raylu, line 363: probably remove this or mention what game we're connecting for if you keep it

src/Game.ts
windo: If you felt like asking, I'm happy to add a note!

src/Game.ts
windo: Whoops, yes, this was a leftover from debugging, will remove!

PR #425, opened 2024-11-22
Update example_config.json5

poonpunokok:


anoek:
This seems to be a mistake

PR #427, opened 2025-02-03
chat: correctly handle player and non-player chats

windo:
Chat messages from players are sent immediately to the client, but non-player chats are not. And then when the game ends, all (player and non-player) chats are sent once again.

So to avoid duplicating chats from the players, keep track of the timestamp of the last chat received, to ignore the duplicates sent at the end of the game.

And to also include chats during the game from non-players, don't filter on the cutoff time for non-player chats - they will always be sent just once anyway.


anoek:
Thanks for the patch!

PR #429, opened 2025-03-22
fix max_time rejection message

siimphh:
I guess this just slipped when all these different failure types were being created!


anoek:
Thanks for the catch and fix!

PR #430, opened 2025-03-22
A few tweaks to pause handling

siimphh:
The first change fixes a bug where the unpause timeout does not trigger.

The second change is to not let players start a "blitz" game but then pause the game and play the entire game paused. I don't think there's any good reason for a user to do that and it can only confuse any stats/limits/logic that might assume that fast games actually are fast.

Third one follows right on forcing fast games to actually be fast - don't include paused games in ongoing game limits on the assumption that the pauses are actually used to walk away from the game so no resources will be used meanwhile.

### DIFF
--- a/src/Game.ts
+++ b/src/Game.ts
@@ -24,6 +24,7 @@ export class Game extends EventEmitter<Events> {
 
     game_id: number;
     state: GobanEngineConfig;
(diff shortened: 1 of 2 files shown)


anoek:
Good improvements, thanks!

PR #434, opened 2026-07-11
Fix checkAllowedRank to use the challenger's rank

GreenAsJade:
## Summary

This change passes `notification.user?.ranking || 0` — the actual challenger's rank, with `|| 0` failing safe so a challenger without rank data is rejected (as "too low") to checkAllowedRank, and adds tests for this and closely related.

( https://forums.online-go.com/t/possible-critical-bug-in-gtp2ogs-challenger-rank-restrictions-being-bypassed/60416?u=greenasjade )


anoek:
Thanks!

PR #435, opened 2026-09-22
Update dependencies to resolve Dependabot security alerts

roy7:
Resolves 45 of the 46 open Dependabot alerts. Only `package.json` and `yarn.lock` change; no source changes.

## Changes

- **`ws` ^8.13.0 → ^8.21.0** — the only runtime change (resolves to 8.21.3). Fixes the high (memory exhaustion DoS) and medium (uninitialized memory disclosure) alerts. Stays within 8.x. No other runtime dependency changes version.
- **`webpack` ^5.76.3 → ^5.104.1**
- **`gulp` 4 → 5** — gulp 4's chokidar 2 / micromatch 3 chain was the source of the old `braces` and `decode-uri-component` versions. The Gulpfile only uses `task`/`series`/`parallel`/`watch`/`src`; `gulp schema` and `gulp eslint` run fine and the regenerated schema is identical.
- **Removed unused dev dependencies:**
  - `gulp-sourcemaps`, `gulp-typescript`: only referenced in commented-out code in `Gulpfile.js` (`gulp-sourcemaps` pulled in `postcss@7`)
  - `prettier-eslint`: not referenced by any script, config or editor setting (it pulled in an exact-pinned vulnerable `minimatch@9.0.3`)
- All other alerts are transitive dev dependencies (eslint/jest/typedoc/etc. tooling), re-resolved to patched versions within existing ranges. The lockfile was updated only for vulnerable packages, not refreshed as a whole.

## Not fixed

- **`pkg` (medium):** no patched release exists (the project is archived). It's still referenced by the `pkg` section of `package.json` for building standalone executables, so I left it alone. Removing it or switching to Node's single-executable applications is your call.

## Testing

- `yarn build`, `yarn lint`, `yarn prettier:check`, `yarn test` (16/16) pass on Node 24 and on Node 18 (matching CI)
- Checked every alert's vulnerable range against the resolved lockfile versions: all 46 match on the old lockfile, only `pkg` on the new one
- **Not tested against a live game.** Given the `ws` bump is a minor release within 8.x the risk seems low, but worth a quick run on beta before release.

## CI note

`.github/workflows/main.yml` only triggers on push/PR to `devel`, so CI won't run for this PR (or other PRs against `main`) now that `main` is the default branch. Probably just needs `main` added to the branch lists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


roy7:
Putting my AI bots to use. ;)

